PR #44, opened 2013-07-18
Initialize sip before importing PyQt.

jwiggins:


LunarLanding:
Well, I don't know how my installation is non - standard enough to suffer of this, but if a simple line re-ordering could stop these problems from consuming time of some people, it would be great.

Issue #108, opened 2013-12-04
message_box example and factories(about, critical, questions, etc.) don't work


blakejohnson:
In trying to run the message_box example, if I click on any of the buttons I get the error message: `AttributeError: 'NoneType' object has no attribute 'setWidget'`. The call stack shows that this happens in the call to `set_parent`.

Possibly relevant: I am using PySide rather than PyQt.

kiwi0fruit:
Message box doesn't work with **PySide2**

Simply using message box leads to error with PySide2: 

```py
from enaml.stdlib.message_box import information
...
    triggered ::
        information(self, "Testing",
            ("If you see this then presumably PySide-Enaml bug was patched.\n" +
             "But it still can be broken if Python hangs after pressing OK.")
        )
...
```
Error:
```
Traceback (most recent call last):
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\qt\qt_action.py", line 79, in on_triggered
    self.declaration.triggered(checked)
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\core\declarative_meta.py", line 68, in declarative_change_handler
    engine.write(owner, change['name'], change)
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\core\expression_engine.py", line 217, in write
    pair.writer(owner, name, change)
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\core\standard_handlers.py", line 82, in __call__
    call_func(func, (), {}, scope)
  File "C:\PyCharmProjects\enaml_video\enaml_video_view.enaml", line 51, in f
    )
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\stdlib\message_box.enaml", line 148, in information
    return _exec_box(parent, title, text, buttons, 'information')
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\stdlib\message_box.enaml", line 305, in _exec_box
    box.set_parent(parent)
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\core\object.py", line 160, in set_parent
    parent.child_added(self)
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\widgets\toolkit_object.py", line 185, in child_added
    child.activate_proxy()
  File "C:\Miniconda3\envs\enaml\lib\site-packages\enaml\widgets\toolkit_object.py", line 210, in activate_proxy
    self.activate_top_down()
[...]

PR #111, opened 2013-12-09
Work around some mro issues with Pyside and QDialog

caryan:
This "fixes" nucleic/enaml#108.  It seems a hack to me but the mro issues are beyond my abilities.  My understanding of the issue is
- QWindowDialog inherits from both QDialog and QWindowBase.  This gives a diamond inheritance pattern at QWidget but I'm not sure if that's the problem. 
- The layout setup happens in `QWindowBase.__init__`
- In `PyQt: QDialog.__init__` is a `<slot wrapper '__init__' of 'sip.simplewrapper' objects>` but `somehow the QWindowBase __init__` is called. 
- In `PySide: QDialog.__init__` is a `<slot wrapper '__init__' of 'PySide.QtGui.QDialog' objects>` and `QWindowBase.__init__` is not called. 

Trying in PySide to only call QWindowBase.**init** led to seg. faults.

### DIFF
--- a/enaml/qt/qt_dialog.py
+++ b/enaml/qt/qt_dialog.py
@@ -9,12 +9,13 @@ from atom.api import Typed
 
 from enaml.widgets.dialog import ProxyDialog
 
-from .QtCore import Qt, Signal
-from .QtGui import QDialog
+from .QtCore import Qt, Signal, QSize
+from .QtGui import QDialog, QLayout
 
-from .q_window_base import QWindowBase
+from .q_window_base import QWindowBase, QWindowLayout
 from .qt_window import QtWindow
 
+from . import QT_API 
 
 class QWindowDialog(QDialog, QWindowBase):
     """ A window base subclass which implements dialog behavior.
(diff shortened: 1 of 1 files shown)


frmdstryr:
I also have another hack to fix this here: https://github.com/frmdstryr/enaml/commit/c9cebd72ac6152217ac8a0689a882576c240ba31

kiwi0fruit:
Both patches proposed ([1st](https://github.com/nucleic/enaml/pull/111/commits/f5784849ec0b47bbffa978569fe6723523ece61e) and [2nd](https://github.com/frmdstryr/enaml/commit/c9cebd72ac6152217ac8a0689a882576c240ba31)) do not fix the problem on Enaml 0.10.0 (Python 3 version from `conda install -c ecpy enaml`).

Windows 7 x64 Miniconda python. Checked with Python 3.6 x64, PySide from `conda install -c conda-forge pyside` and with Python 3.4 win32, Pyside from `pip install pyside`.

kiwi0fruit:
Additioanl info: these two patches also do not work on Enaml 0.10.0 at Python 2.7 (same behaviour as described above). But they do work on Enaml 0.9.8 at Python 2.7.

This may indicate that some instability was introduced to Enaml python 3 version (Error code 0xc0000005 means access violation).

MatthieuDartiailh:
Closed by #423

PR #160, opened 2014-07-13
Add completions to field with Qt implementation

blink1073:
This is in response to a Google-Groups [discussion](https://groups.google.com/forum/#!topic/enaml/WMHkQOMifqg).  It adds a `completer` attribute to a `Field` and an associated `Completer` object.  The default `Completer` has a list of completions.  I added a `FileCompleter` and a `DirectoryCompleter` that handle system paths, and a `CustomCompleter` that takes a callback function to update the completion list.  I believe this covers the use cases we discussed.  Interestingly, this could be added to the `ComboBox`, which also supports completers.  As before, there is only a Qt implementation, but [wx](http://docs.wxwidgets.org/trunk/classwx_text_entry.html) does support this construct.

I used the following code to test:

``` python
from enaml.widgets.api import Window, Form, Label, Field
from enaml.completer import Completer

COMPLETIONS='''Germany;Russia;France;
french fries;frizzy hair;fennel;fuzzball
frayed;fickle;Frobozz;fear;framing;frames
Franco-American;Frames;fancy;fire;frozen yogurt
football;fnord;foul;fowl;foo;bar;baz;quux
family;Fozzie Bear;flinch;fizzy;famous;fellow
friend;fog;foil;far;flower;flour;Florida
'''.replace('\n',';').split(";")

enamldef Main(Window):
    Form:
        Label:
            text = 'First Name'
        Field:
            completer=Completer(completions=COMPLETIONS)
```


blink1073:
Closing as stale.

PR #179, opened 2015-02-19
Fixes for font handling

blink1073:
Make font parsing more flexible to order.  These now both work: `arial 12pt bold`, `12pt arial bold`.

Also corrected example that uses unsupported bare integer font size.

### DIFF
--- a/enaml/fonts.py
+++ b/enaml/fonts.py
@@ -183,7 +183,8 @@ class FontMember(Coerced):
 
     A font member can be set to a Font, a string, or None. A string
     font will be parsed into a Font object. If the parsing fails,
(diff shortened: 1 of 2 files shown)


sccolbert:
The parser follows the CSS grammar, which is order specific:

http://www.w3.org/TR/REC-CSS1/#font

`bold 12pt arial` is the only valid permutation according to the CSS grammar.

Qt may support other permutations, but allowing that will require extra work to support to another backend like HTML. I'm not sure its worth the tradeoff.

blink1073:
How does that look?

sccolbert:
Looks good to me. Thanks!

PR #180, opened 2015-02-19
Fix mpl deprecation warning and add ability to supply canvas/toolbar

blink1073:
The `NavigationToolbar2QTAgg` has been deprecated.
Also, I have an application that uses a custom Canvas and Toolbar, and this allows me to use them.
The `canvas.manager.toolbar = None` prevents the `FigureManager` from trying to destroy the toolbar, which resulted in a `RuntimeError`.

### DIFF
--- a/enaml/qt/qt_mpl_canvas.py
+++ b/enaml/qt/qt_mpl_canvas.py
@@ -99,13 +99,13 @@ class QtMPLCanvas(QtControl, ProxyMPLCanvas):
                 layout.addWidget(toolbar)
                 layout.addWidget(self.canvas)
                 self.canvas.setParent(widget)
-                # use focus policy from mpl FigureManager
+                # Use focus policy from MPL FigureManager
                 self.canvas.setFocusPolicy(Qt.StrongFocus)
                 self.canvas.setFocus()
                 self.canvas.setVisible(True)
             else:
                 canvas = self.canvas
-                # reset and clear the toolbar
+                # Reset and clear the toolbar
                 canvas.toolbar.home()
                 canvas.toolbar.update()
                 figure.canvas = canvas
--- a/enaml/qt/qt_mpl_canvas.py
+++ b/enaml/qt/qt_mpl_canvas.py
@@ -25,6 +25,9 @@ class QtMPLCanvas(QtControl, ProxyMPLCanvas):
     #: A reference to the widget created by the proxy.
     widget = Typed(QFrame)
 
+    #: MPL Canvas Object (Singleton)
+    canvas = Typed(FigureCanvasQTAgg)
+
     #--------------------------------------------------------------------------
     # Initialization API
     #--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


sccolbert:
I'm worried about the code in the block above:
https://github.com/blink1073/enaml/blob/mpl-deprecation/enaml/qt/qt_mpl_canvas.py#L76

I haven't fully grok'd your changes yet, but my comments seem to indicate it was easier to bail and re-create everything, than trying to reuse the old figure widgets. Hence, the calls to deleteLater. Attempting to reuse `figure.canvas` is bound to lead to problems, since it will be deleted at some point in the future.

Have you tested this change against swapping the figure?

blink1073:
No, I tried it with an existing `FigureCanvasQTAgg`.  I'll dig into it.

blink1073:
Tested against everything I could think of to throw at it: existing canvas/new canvas, toolbar toggling, zooming.  It switches seamlessly between the two figures.  I also used the focus defaults from `FigureManager` that solved the key press issue.

enaml/qt/qt_mpl_canvas.py
```diff
@@ -89,11 +89,17 @@ def _refresh_mpl_widget(self):
         # visible, or QVBoxLayout will ignore it for size hinting.
         figure = self.declaration.figure
         if figure:
-            canvas = FigureCanvasQTAgg(figure)
+            if isinstance(figure.canvas, FigureCanvasQTAgg):
+                canvas = figure.canvas
+                canvas.manager.toolbar = None
+                toolbar = canvas.toolbar
+                toolbar.setParent(widget)
```
sccolbert: This call to `setParent` is not needed. The layout will automatically reparent the widget when it's added to the layout.

enaml/qt/qt_mpl_canvas.py
```diff
@@ -89,11 +89,17 @@ def _refresh_mpl_widget(self):
...
+            else:
+                canvas = FigureCanvasQTAgg(figure)
+                toolbar = NavigationToolbar2QT(canvas, widget)
             canvas.setParent(widget)
```
sccolbert: `setParent` not needed

enaml/qt/qt_mpl_canvas.py
```diff
@@ -89,11 +89,17 @@ def _refresh_mpl_widget(self):
         # visible, or QVBoxLayout will ignore it for size hinting.
         figure = self.declaration.figure
         if figure:
```
sccolbert: What happens if there is no `figure` (i.e. it was set to None in Enaml)? Does the old stale figure continue to be displayed?

sccolbert:
ugh, i dont know why i was pointed at the outdated diff. I will repost those comments.

enaml/qt/qt_mpl_canvas.py
```diff
@@ -73,27 +76,48 @@ def _refresh_mpl_widget(self):
...
-        # which is certainly not desired in this case. This appears to
-        # be a limitation of matplotlib. The canvas is manually set to
-        # visible, or QVBoxLayout will ignore it for size hinting.
+
+        # Create the new canvas and toolbar widgets if suitable ones have
+        # not been created.
+        # The canvas is manually set to visible, or QVBoxLayout will
+        # ignore it for size hinting.
         figure = self.declaration.figure
         if figure:
```
sccolbert: What happens if the `figure` has been set to `None` in Enaml? Is the old stale figure still displayed? If so, that should be fixed.

enaml/qt/qt_mpl_canvas.py
```diff
@@ -73,27 +76,48 @@ def _refresh_mpl_widget(self):
...
+                    canvas = figure.canvas
+                else:
+                    canvas = FigureCanvasQTAgg(figure)
+
+                if canvas.toolbar:
+                    # Avoid RuntimeError due to multiple calls to destroy
+                    if hasattr(canvas, 'manager'):
+                        canvas.manager.toolbar = None
+                    toolbar = canvas.toolbar
+                    toolbar.setParent(widget)
```
sccolbert: This call to `setParent` is not needed. The layout will automatically reparent the widget when it's added to the layout.

enaml/qt/qt_mpl_canvas.py
```diff
@@ -73,27 +76,48 @@ def _refresh_mpl_widget(self):
...
+                else:
+                    toolbar = NavigationToolbar2QT(canvas, widget)
+
+                layout.addWidget(toolbar)
+                layout.addWidget(canvas)
+                canvas.setParent(widget)
```
sccolbert: `setParent` is not needed here

enaml/qt/qt_mpl_canvas.py
```diff
@@ -73,27 +76,48 @@ def _refresh_mpl_widget(self):
...
+                # Use focus policy from MPL FigureManager
+                canvas.setFocusPolicy(Qt.StrongFocus)
+                canvas.setFocus()
```
sccolbert: I'm don't think we should be stealing focus here. If the user wants to set focus on the widget, they should call the `set_focus` method on the Enaml widget:
https://github.com/blink1073/enaml/blob/mpl-deprecation/enaml/widgets/widget.py#L191

To make sure the canvas gets focus when that happens, you can reimplement `focus_target` to return the canvas:
https://github.com/blink1073/enaml/blob/mpl-deprecation/enaml/qt/qt_widget.py#L184

blink1073:
Now the canvas is invisible (and does not accept events) when set to `None`.  I believe I've hit all your concerns.

sccolbert:
I made a few changes. Can you try this branch and check to make sure it handles your use case?
https://github.com/nucleic/enaml/tree/blink1073-mpl-deprecation

blink1073:
Verified that it all works, thanks!  I forgot another use case for this.  You don't have to wait until after `enaml` creates a canvas to attach events to it.

blink1073:
I spoke too soon, when I switch figures the `key_press` callback does not work.

blink1073:
Nevermind, what you have now makes more sense, since each figure carries with it its original canvas.

blink1073:
Closing as stale.

Issue #172, opened 2014-11-19
IntField handles only Int(), not Long(), fails in cryptic way


jason-s:
I'm working with unsigned 32-bit integers, which won't fit in an `int`, so I need to use an `atom.api.Long` rather than an `atom.api.Int`, but then `IntField` won't work. But there is no `LongField` or `LongValidator` :/

This gist https://gist.github.com/jason-s/5f884e80ffd8a7c15af7 fails with `TypeError: invalid instance type` when I use enaml-run. 

```
C:\mydir>enaml-run testdev/test1.enaml
Traceback (most recent call last):
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\Scripts\enaml-run-script.py",
line 5, in <module>
    sys.exit(main())
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\runner
.py", line 57, in main
    exec code in ns
  File "testdev/test1.enaml", line 30, in
    main()
  File "testdev/test1.enaml", line 23, in main
    view.show()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\widget
s\window.py", line 377, in show
    self.activate_proxy()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\widget
s\toolkit_object.py", line 206, in activate_proxy
    child.activate_proxy()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\widget
s\toolkit_object.py", line 206, in activate_proxy
    child.activate_proxy()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\widget
s\toolkit_object.py", line 203, in activate_proxy
    self.activate_top_down()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\widget
s\toolkit_object.py", line 219, in activate_top_down
    self.proxy.activate_top_down()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\qt\qt_
toolkit_object.py", line 76, in activate_top_down
    self.init_widget()
  File "c:\app\python\anaconda\1.6.0\envs\emblaze\lib\site-packages\enaml\qt\qt_
field.py", line 69, in init_widget
    if d.text:
[...]

PR #181, opened 2015-04-01
ComplexField

AndyGreenwell:
Pull request for adding a ComplexField enamldef to fields.enaml for the purpose of input and display of complex numbers as well as a corresponding ComplexValidator within validator.py.  The content was created by copying and slightly modifying the FloatField enamldef and FloatValidator class.

### DIFF
--- a/enaml/validator.py
+++ b/enaml/validator.py
@@ -151,6 +151,57 @@ class FloatValidator(Validator):
         return True
 
 
+class ComplexValidator(Validator):
+    """ A concrete Validator which handles complex floating point input.
+
+    This validator ensures that the text represents a complex floating point
+    number within a specified range.
+
+    """
+    #: The minimum value allowed for the float, inclusive, or None if
+    #: there is no lower bound.
+    minimum = Typed(float)
+
+    #: The maximum value allowed for the float, inclusive, or None if
+    #: there is no upper bound.
+    maximum = Typed(float)
+
+    #: Whether or not to allow exponents like '1e6' in the input.
+    allow_exponent = Bool(True)
+
+    def validate(self, text):
+        """ Validates the given text matches the complex float range.
+
+        Parameters
+        ----------
+        text : unicode
+            The unicode text edited by the client widget.
+
+        Returns
+        -------
+        result : bool
+            True if the text is valid, False otherwise.
+
+        """
+        try:
+            value = complex(text)
+        except ValueError:
+            return False
+        minimum = self.minimum
+        if minimum is not None and (value.real < minimum or value.imag < minimum): 
+			return False
+			
+        maximum = self.maximum
+		
+        if maximum is not None and (value.real > maximum or value.imag > maximum):
+            return False
+			
+        if not self.allow_exponent and 'e' in text.lower():
+            return False
+			
+        return True
+        
+
 class RegexValidator(Validator):
     """ A concrete Validator which handles text input.
 
--- a/enaml/stdlib/fields.enaml
+++ b/enaml/stdlib/fields.enaml
@@ -10,7 +10,7 @@
 This is a library of Enaml components deriving from the builtin Field.
 
 """
-from enaml.validator import RegexValidator, IntValidator, FloatValidator
+from enaml.validator import RegexValidator, IntValidator, FloatValidator, ComplexValidator
 from enaml.widgets.field import Field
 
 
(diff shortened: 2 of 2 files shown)


enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+class ComplexValidator(Validator):
+    """ A concrete Validator which handles complex floating point input.
+
+    This validator ensures that the text represents a complex floating point
+    number within a specified range.
+
+    """
+    #: The minimum value allowed for the float, inclusive, or None if
+    #: there is no lower bound.
+    minimum = Typed(float)
```
sccolbert: should be Typed(complex)

enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+
+    #: The maximum value allowed for the float, inclusive, or None if
+    #: there is no upper bound.
+    maximum = Typed(float)
```
sccolbert: should be `Typed(complex)`

enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+            True if the text is valid, False otherwise.
+
+        """
+        try:
+            value = complex(text)
+        except ValueError:
+            return False
+        minimum = self.minimum
+        if minimum is not None and (value.real < minimum or value.imag < minimum): 
+			return False
```
sccolbert: looks like tabs for indentation on this line - should be 4 spaces

enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+
```
sccolbert: no blank line

enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+        try:
+            value = complex(text)
+        except ValueError:
+            return False
+        minimum = self.minimum
+        if minimum is not None and (value.real < minimum or value.imag < minimum): 
+			return False
+			
+        maximum = self.maximum
+
```
sccolbert: no blank line

enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+            return False
+        minimum = self.minimum
+        if minimum is not None and (value.real < minimum or value.imag < minimum): 
+			return False
+			
+        maximum = self.maximum
+		
+        if maximum is not None and (value.real > maximum or value.imag > maximum):
+            return False
+
```
sccolbert: no blank line

enaml/validator.py
```diff
@@ -151,6 +151,57 @@ def validate(self, text):
...
+			return False
+			
+        maximum = self.maximum
+		
+        if maximum is not None and (value.real > maximum or value.imag > maximum):
+            return False
+			
+        if not self.allow_exponent and 'e' in text.lower():
+            return False
+
```
sccolbert: no blank line

sccolbert:
Looks mostly g2g. I made a few formatting comments and have a question about having the min/max bounds be floats instead of complex. Was there a particular reason for that?

AndyGreenwell:
No particular reason for the choices made other than just performing the minimal changes needed from the existing FloatField and FloatValidator to get a minimal level of functionality working.  With min/max check, mirroring the numpy min/max behavior is probably better. I planned to submit the pull request and then wait for review and corrections.  Looks like that plan worked. :)

Currently working on a project that needed input and display of complex numbers, and I really like the library! Thanks,.

jason-sachs:
Hmm. Mixed thoughts, this is really specialized behavior, I don't have any strong objection to a ComplexField object, it just seems a bit odd for a general framework. I had a [similar request](https://github.com/nucleic/enaml/issues/172) for a LongField.

AndyGreenwell:
When building applications for physics or engineering, the display and
entry of complex numbers is a common use case.  That is my motivation here,
as it is functionality that was missing for my current needs.

One can always break up the complex number into two components and display
the real and imaginary parts in separate fields, but that introduces
(likely unnecessary) extra operations on the
domain-specialist's/application-developer's part when their application is
processing complex numbers.

If the maintainers decide that this functionality is not necessary, then
please deny this pull request, and I will move forward with a different
implementation in my current application.

Thanks,

Andy

sccolbert:
@AndyGreenwell I'm all for the pull request. Please update your PR to address my comments and then I'll merge it.

@jason-sachs The difference between this PR and what you were suggesting is that there is no long type in Python 3, and Python 3 will be the target going forward.

sccolbert:
Actually, does it even make sense to have a range on complex numbers? There's not really a standard for relative ordering of complex numbers...

blink1073:
Well, you could constrain on magnitude and/or phase...

AndyGreenwell:
True, complex numbers are not an ordered field, so just imposing an arbitrary order in code might cover some cases cleanly but not others.

That said, being able to impose minimum and maximum values on the real and imaginary parts separately can be useful in some physical contexts.  For example, in an electrical engineering context, one might have separate limits on the real and imaginary parts of the impedance (resistance and reactance), but end up doing computations on the complex quantity.  While that example may in some ways be arguing against inclusion of a ComplexField, having a field that neatly accepts, displays, and validates a complex value is useful.

I defer to your better judgment on whether this functionality should be included.

sccolbert:
Instead of having min/max range on this validator, what about a callable property which the user could assign which would validate the value after it passes the initial string -> complex conversion?

PR #182, opened 2015-04-03
Declarative function examples

MatthieuDartiailh:
This adds some examples on how to use declarative functions. Some examples are not very inspired but at least they show how things work. I would like to add a fourth one so this is not ready to merge. But feel free to review what is already there.

### DIFF
--- a/examples/functions/declare_function.enaml
+++ b/examples/functions/declare_function.enaml
@@ -1,0 +1,43 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+""" An example showing how to declare a declarive function using func
+
+This example demonstrates how one can use the `func` keyword to declare
+a function bound to an enaml declaration and how scoping rules apply.
+
+Note that when declaring a function using the `func` it will always be
+considered a method of the object on which it is defined hence the possibility
+to access to self which is a reference to the main window.
+
+Note also that such a function has access to the same scope as any other Enaml
+object allowing for example to access the SpinBox using its id which would not
+be possible using a function (defined using def) declared outside the enamldef
+block.
+
+<< autodoc-me >>
+"""
+from enaml.widgets.api import Window, PushButton, SpinBox, Container
+from enaml.layout.api import hbox, vbox
+
+
+enamldef Main(Window):
+    attr thing = 'twelve'
+    attr thing2 = 'nine'
+
+    func do_thing(param):
+       print thing, self.thing2, sbox.value, param
+
+    Container:
+
+        SpinBox: sbox:
+            maximum = 100
+            minimum = 0
+
+        PushButton:
+            text = 'Click Me'
+            clicked :: do_thing(42)
--- a/examples/functions/declare_function.enaml
+++ b/examples/functions/declare_function.enaml
@@ -12,7 +12,7 @@ a function bound to an enaml declaration and how scoping rules apply.
 
 Note that when declaring a function using the `func` it will always be
 considered a method of the object on which it is defined hence the possibility
-to access to self which is a reference to the main window.
+to access self which in this case is a reference to the main window.
 
 Note also that such a function has access to the same scope as any other Enaml
 object allowing for example to access the SpinBox using its id which would not
@@ -22,7 +22,6 @@ block.
 << autodoc-me >>
 """
 from enaml.widgets.api import Window, PushButton, SpinBox, Container
-from enaml.layout.api import hbox, vbox
 
 
 enamldef Main(Window):
--- a/examples/functions/observe_model_signal.enaml
+++ b/examples/functions/observe_model_signal.enaml
@@ -1,0 +1,125 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+""" An example showing how to use to observe a model signal.
+
+Here the model use a signal to notify listeners about in place change to a
+list. This pattern is interestign when a ContainerList would emit too
+many signal because notifications cannot be batched. This is a common case
+when moving an element in a list for example.
+
+<< autodoc-me >>
+"""
+from atom.api import Atom, List, Signal
+from enaml.widgets.api import (MainWindow, Form, PushButton, Field,
+                               Menu, Action)
+from enaml.core.api import Looper
+
+
+class Model(Atom):
+    """Simple model holding simply a list.
+
+    This model expose a list which should be considered read-only and methods
+    to manipulate it. Those methods use a signal to notify listeners that
+    inplace changes occured on the list.
+
+    """
+    values = List(default=['val'])
+
+    values_changed = Signal()
+
+    def add_value(self, index, val):
+        """Add a value at a specified index.
+
+        """
+        self.values.insert(index, val)
+        self.values_changed('add')
+
+    def move_value(self, old, new):
+        """Move a value from one index to another.
+
+        """
+        val = self.values.pop(old)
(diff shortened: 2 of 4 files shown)


examples/layout/advanced/override_layout_constraints.enaml
```diff
@@ -0,0 +1,58 @@
...
+<< autodoc-me >>
+"""
+from itertools import izip_longest
+
+from enaml.core.api import Include
+from enaml.layout.api import align, grid, factory
+from enaml.widgets.api import Window, Container, Field, Label, PushButton
+
+
+def generate_grid():
```
blink1073: Is this function left over?

MatthieuDartiailh: Indeed I will remove it when adding the last example.

jason-s:
dumb question: why use a special keyword `func` rather than Python's `def`?

```
func do_thing(param):
  print thing, self.thing2, sbox.value, param
```
vs.

```
def do_thing(self, param):
  print self.thing, self.thing2, self.sbox.value, param
```

sccolbert:
Two reasons:

1) It makes it easier to parse
2) A `func` does not behave like a normal function. It has a magic `self` and dynamic scoping.

examples/functions/declare_function.enaml
```diff
@@ -0,0 +1,43 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+""" An example showing how to declare a declarive function using func
+
+This example demonstrates how one can use the `func` keyword to declare
+a function bound to an enaml declaration and how scoping rules apply.
+
+Note that when declaring a function using the `func` it will always be
+considered a method of the object on which it is defined hence the possibility
+to access to self which is a reference to the main window.
```
sccolbert: can you change this to "to access `self` which in this case is a reference to the main window".

examples/layout/advanced/override_layout_constraints.enaml
```diff
@@ -0,0 +1,58 @@
...
+    """
+    rows = []
+    widgets = self.visible_widgets()
+    row_iters = (iter(widgets),) * 2
+    rows = list(izip_longest(*row_iters))
+    return [grid(*rows), align('width', *widgets)]
+
+
+enamldef Main(Window):
+    title = 'Factory Helper'
```
sccolbert: this title doesn't make sense with this example

examples/layout/advanced/override_layout_constraints.enaml
```diff
@@ -0,0 +1,58 @@
...
+            rows = list(izip_longest(*row_iters))
+            return [grid(*rows)] + [align('v_center', *row) for row in rows]
+
+        Label:
+            text = 'Name'
+        Field:
+            pass
+        Label:
+            text = 'Surname'
+        Field: c_count:
```
sccolbert: this id isn't used, should probably be removed

examples/layout/advanced/override_layout_constraints.enaml
```diff
@@ -0,0 +1,58 @@
...
+different from the deafults ones.
+
+<< autodoc-me >>
+"""
+from itertools import izip_longest
+
+from enaml.core.api import Include
+from enaml.layout.api import align, grid, factory
+from enaml.widgets.api import Window, Container, Field, Label, PushButton
+
```
sccolbert: cleanup unused imports

examples/functions/override_function.enaml
```diff
@@ -0,0 +1,48 @@
...
+Note that when subclassing an enaml object you must use the @d_func decoarator
+to mark a function as declarative and hence overridable from enaml.
+
+<< autodoc-me >>
+"""
+from atom.api import observe
+from enaml.core.declarative import d_func
+from enaml.widgets.api import Window, PushButton, Container
+from enaml.layout.api import hbox, vbox
+
```
sccolbert: cleanup unused imports

examples/functions/declare_function.enaml
```diff
@@ -0,0 +1,43 @@
...
+
+Note also that such a function has access to the same scope as any other Enaml
+object allowing for example to access the SpinBox using its id which would not
+be possible using a function (defined using def) declared outside the enamldef
+block.
+
+<< autodoc-me >>
+"""
+from enaml.widgets.api import Window, PushButton, SpinBox, Container
+from enaml.layout.api import hbox, vbox
```
sccolbert: cleanup unused imports

MatthieuDartiailh:
Comments addressed and last example added.
This last example is a bit more complex and I don't know if it presents the right approach to achieve what it tries to achieve, but it does work and seems nice to me so I thought it would be nice to share it.

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,125 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+""" An example showing how to use to observe a model signal.
+
+Here the model use a signal to notify listeners about in place change to a
```
sccolbert: s/use/uses

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,125 @@
...
+list. This pattern is interestign when a ContainerList would emit too
```
sccolbert: s/interestign/interesting

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,125 @@
...
+many signal because notifications cannot be batched. This is a common case
```
sccolbert: s/signal/signals

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,125 @@
...
+    def delete_value(self, index):
+        """Delete a value.
+
+        """
+        del self.values[index]
+        self.values_changed('deleted')
+
+
+enamldef EditMenu(Menu):
+    """Menu use to edit the content of the list.
```
sccolbert: s/use/used

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,125 @@
...
+
+    The entries depend on the state of the list to avoid proposing absurd
```
sccolbert: This line is a bit unclear. Can you rephrase?

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,125 @@
...
+enamldef Main(MainWindow): main:
+
+    attr model = Model()
+    attr _internal << model.values[:]
+
+    Form:
+        func react_signal(kind):
+            main._internal = model.values[:]
+            if kind == 'moved':
+                self.request_relayout()
```
sccolbert: is this request layout needed?

sccolbert:
Looks pretty good. A few more comments on the new example.

MatthieuDartiailh:
Comments addressed. I added a comment about the request_relayout. I must say I am not sure to understand why it works but it does. One thing that I do not understand is why the order of the children is respected but the Looper does not update their loop_index. It might be something to fix.

examples/functions/observe_model_signal.enaml
```diff
@@ -0,0 +1,133 @@
...
+
+enamldef Main(MainWindow): main:
+
+    attr model = Model()
+    attr _internal << model.values[:]
+
+    Form:
+        func react_signal(kind):
+            main._internal = model.values[:]
+            if kind == 'moved':
```
sccolbert: This whole "if" block can be removed. I just pushed an update which fixes the lack of automatic relayout when children are simply moved instead of added/removed.
https://github.com/nucleic/enaml/commit/e0a886a899deb45bb4c255f8fe400fa5db0ca19c

sccolbert:
Just fixed the bug for the relayout request and made a comment on one last thing to fix before merging.

MatthieuDartiailh:
Last comment addressed !

sccolbert:
Thanks!

PR #187, opened 2015-06-24
Added entry point for extensions to add custom factories

gabrielcnr:
Hey Chris, how are you?

Do you still remember me? :-)

So, I'm proposing this simple change that will already be enough as an extension/entry point to allow enaml plugins to declare custom factories.

I also created a simple demo project that makes usage of the feature in:
https://github.com/gabrielcnr/enaml_hello

Can you please take a look to see if this can be merged?

Thanks,
Gabe


sccolbert:
This is an interesting idea :)

One thing I think of though, is that users of the extension will still need to import the Enaml declaration object. The extension could simpy register its factories during the import. What does this proposal gain us over that approach?

gabrielcnr:
I'm not sure if I get exactly what you mean.

Are you suggesting that, on the .enaml module, instead of doing

`from path.to.custom_control import CustomControl`

you would simply write:

`from enaml.widgets.api import CustomControl` ?

If that's what you mean, I had kind of thought about that, but I wasn't sure how to roll that out without breaking the lazy import paradigm for the Qt Controls.
Anyway, I made the changes and pushed here so we can discuss.

sccolbert:
That's not quite what I meant. You still have your custom module, but at the bottom of that module you register your factories. So, when the user imports your module, your factories get automatically registered. Since factories are callable, you can still delay any toolkit specific imports until the factory is called.

What you are proposing is actually a pretty big conceptual change, which I'm hesitant to make if there is another way to solve your problem with the existing framework.

gabrielcnr:
Right, I think I see now what you mean. But I believe now that I'm missing here some fundamental piece of knowledge. Is there another way to register factories but the one via adding keys to the QT_FACTORIES dict?

Or are you suggesting to put something like...

```python
def hello_factory():
    from enaml_hello.qt.qt_hello import QtHello
    return QtHello

from enaml.qt.qt_factories import QT_FACTORIES
QT_FACTORIES['Hello'] = hello_factory
```

...at the bottom of my module where I define my `class Hello(Control)`?

Thanks

sccolbert:
Yes, that's what I'm suggesting.

MatthieuDartiailh:
Closing as there is a working alternative.

PR #189, opened 2015-06-29
MNT: Matplotlib deprecated NavigationToolbar2QTAgg

ericdill:


ericdill:
With matplotlib master, `NavigationToolbar2QTAgg` no longer exists. It has been moved to `NavigationToolbar2QT`

sccolbert:
Is it just a simple renaming in matplotblib, or did the API change as well?

blink1073:
It has a deprecation warning right now, will be removed in a future release.

sccolbert:
Right, but do we to change our use of the object? i.e. is NavigationToolbar2QT fully compatible in both API and behavior as NavigationToolbar2QTAgg?

blink1073:
It is functionally the same object, they shuffled the logic back into the main toolbar.

ericdill:
According to @tacaswell 's commit that removed `NavigationToolbar2QTAgg`, there is no additional functionality of `NavigationToolbar2QTAgg` over the base `NavigationToolbar2QT`

see https://github.com/matplotlib/matplotlib/commit/056ba7a87e5bbd37dbf61a3c27329dc4f5648a54

ericdill:
Sorry, the title should say mpl deprecated QTAgg and now it is removed from master

enaml/qt/qt_mpl_canvas.py
```diff
@@ -10,7 +10,10 @@
 from enaml.widgets.mpl_canvas import ProxyMPLCanvas
 
 from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg
-from matplotlib.backends.backend_qt4agg import NavigationToolbar2QTAgg
+try:
+    from matplotlib.backends.backend_qt4agg import NavigationToolbar2QTAgg
```
sccolbert: should be `import NavigationToolbar2QTAgg as NavigationToolbar2QT`

enaml/qt/qt_mpl_canvas.py
```diff
@@ -10,7 +10,10 @@
...
+except ImportError:
+    from matplotlib.backends.backend_qt4agg import NavigationToolbar2QT
```
blink1073: I think this should be `backend_qt5agg`, where it actually resides:
https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/backends/backend_qt4.py#L35

sccolbert: -1. 

I think those imports are there explicitly because they are Qt4 compatible. Using a Qt5 import would be confusing, since we don't support Qt5.

sccolbert:
Cheers!

PR #191, opened 2015-07-20
Fix uninitialized `editable` attribute

sccolbert:
This fixes a bug where `editable` was not properly initialized by the backend ObjectCombo implementation.


bburan:
This fixed the bug.  Thanks!

sccolbert:
Thanks!

PR #192, opened 2015-07-21
Add a syntax definition for BBEdit/TextWrangler

jwiggins:
To install with TextWrangler, copy to: `~/Library/Application Support/TextWrangler/Language Modules/`
See http://www.barebones.com/support/develop/clm.html for more details.


sccolbert:
Cheers!

PR #198, opened 2015-10-13
Removed legacy dependency on distribute.

gabrielcnr:
distribute is a legacy package.

Should the installation dependency be updated to require setuptools instead?


MatthieuDartiailh:
Addressed in #227

PR #200, opened 2015-11-08
qtobjectcombo : stops timer when calling destroy.

### LINKED ISSUE #199: ObjectCombo : timer can fire after object destruction
The timer in charge of collapsing the refresh requests is not stop when destroying and there is no guard checking that the declaration still exists when actually refreshing. This can cause an attribute error.

I can do a PR but I am not sure whether we can stop reliably the timer in the destroy method or to simply add a check in the refresh_items method.

MatthieuDartiailh:
closes #199

### DIFF
--- a/enaml/qt/qt_object_combo.py
+++ b/enaml/qt/qt_object_combo.py
@@ -96,6 +96,13 @@ class QtObjectCombo(QtControl, ProxyObjectCombo):
         self.set_editable(self.declaration.editable)
         self.widget.currentIndexChanged.connect(self.on_index_changed)
 
+    def destroy(self):
+        """Re-implemented here to take care of stopping the refresh timer.
+
+        """
+        self.refresh_timer.stop()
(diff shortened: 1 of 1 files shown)


enaml/qt/qt_object_combo.py
```diff
@@ -96,6 +96,13 @@ def init_widget(self):
         self.set_editable(self.declaration.editable)
         self.widget.currentIndexChanged.connect(self.on_index_changed)
 
+    def destroy(self):
+        """Re-implemented here to take care of stopping the refresh timer.
```
sccolbert: One space after opening triple quote.

sccolbert:
Thanks! Did this solve the problem for you?

Only one minor whitespace correction.

MatthieuDartiailh:
Yes it did. I was worried about race conditions but it does not seem to be an issue.

sccolbert:
Cheers!

Issue #193, opened 2015-07-22
Use Include/Looper in PluginManifest's


frmdstryr:
I'm trying to create a dynamic workbench submenu using a looper in my plugin's manifest, but it's not working. Any suggestions?

```python

        Looper:
            iterable << ['test.doc','test2.doc']
            ActionItem:
                path = '/file/recent/%s'%loop_index
                label = '%s'%loop_item
                command = 'test.workbench.ui.open'
                parameters = {'path':loop_item}
        

```

I've tried using _refresh_actions() of the UI plugin but that doesn't work. Is there a 'correct' way to do this?

sccolbert:
`ActionItem` has to be a child of an `Extension`, not a `PluginManifest`:
https://github.com/nucleic/enaml/blob/master/examples/workbench/sample_plugin.enaml#L61

frmdstryr:
Thanks for the quick reply. I added it to the example as shown below but it doesn't work.

```python

#------------------------------------------------------------------------------
# Copyright (c) 2013, Nucleic Development Team.
#
# Distributed under the terms of the Modified BSD License.
#
# The full license is in the file COPYING.txt, distributed with this software.
#------------------------------------------------------------------------------
from enaml.workbench.api import Extension, PluginManifest
from enaml.workbench.ui.api import ActionItem, Branding, MenuItem, ItemGroup
from enaml.core.looper import Looper
from enaml.core.include import Include

def first_view_factory(workbench):
    from sample_workspace import SampleWorkspace

    import enaml
    with enaml.imports():
        from first_view import FirstView, FirstManifest

    space = SampleWorkspace()
    space.window_title = 'First View'
    space.content_def = FirstView
    space.manifest_def = FirstManifest
    return space


def second_view_factory(workbench):
    from sample_workspace import SampleWorkspace

    import enaml
    with enaml.imports():
        from second_view import SecondView, SecondManifest

    space = SampleWorkspace()
    space.window_title = 'Second View'
    space.content_def = SecondView
    space.manifest_def = SecondManifest
    return space


def persistent_view_factory(workbench):
    from persistent_workspace import PersistentWorkspace
    space = PersistentWorkspace()
    space.window_title = 'Persistent View'
    return space


enamldef SampleManifest(PluginManifest):
    """ The plugin manifest for the primary example plugin.

    This plugin acts as the entry point for all other plugins in this
    example. It contributes the window branding, default actions, and
    the workspace definitions.

    """
    id = 'sample'
[...]

PR #201, opened 2015-11-16
Fix :  Use Include/Looper in PluginManifest's #193 

MatthieuDartiailh:
This was left unresolved. Except for the unnecessary white spaces it looks good. @frmdstryr perhaps you could fix it so that @sccolbert could merge (if he agrees with me).


frmdstryr:
A test or example would also be helpful

MatthieuDartiailh:
Great ! Any hope you can look at #213 while you are around ?

blink1073:
We are full throttle working on Phosphor and JupyterLab in preparation for the talk at the Scipy conference, might be a while yet.

PR #203, opened 2015-11-16
Remove module from sys.modules when an exception occurs

### LINKED ISSUE #202: Importing a module always populate sys.module even when import fails
I noticed something that seems like a bug to me. When importing an enaml module causing an exception to be raised, the module is always added to sys.modules which is not the standard behavior of python. This means that if importing a module a first time and catching the exception, a second import will not raise but return an empty module.
This has bugged me for some hours when writing tests which were not raising the right error because the module was previously imported. I will open a PR in the coming days.

MatthieuDartiailh:
closes #202 

This fixes the issue for me. Should I add a test for it ?

Note : would you consider making a 0.9.9 release for the small bugs that were fixed over the past months ?

### DIFF
--- a/enaml/core/import_hooks.py
+++ b/enaml/core/import_hooks.py
@@ -129,8 +129,10 @@ class AbstractEnamlImporter(object):
         """
         code, path = self.get_code()
         if fullname in sys.modules:
+            pre_exists = True
             mod = sys.modules[fullname]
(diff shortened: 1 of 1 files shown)


sccolbert:
Thanks for the fix! I'll see if I can find the time to release.

PR #205, opened 2016-05-05
Allow subclassing of FileDialogEx

### LINKED ISSUE #204: Cannot subclass FileDialogEx
I cannot subclass FileDialogEx so that I can pre-set some of the fields (e.g., the file name filters) for a common file type. For example:

    enamldef LayoutFileDialog(FileDialogEx):
        name_filters = ['Workspace layout (*.layout)']
        selected_name_filter = '*.layout'

If I call this:

    LayoutFileDialog.get_save_file_name()

I still get the standard file dialog with no filters.

bburan:
This fixes #204

### DIFF
--- a/enaml/widgets/file_dialog_ex.py
+++ b/enaml/widgets/file_dialog_ex.py
@@ -75,8 +75,8 @@ class FileDialogEx(ToolkitDialog):
     #: A reference to the ProxyFileDialog object.
     proxy = Typed(ProxyFileDialogEx)
 
-    @staticmethod
-    def get_existing_directory(parent=None, **kwargs):
+    @classmethod
+    def get_existing_directory(cls, parent=None, **kwargs):
         """ Get an existing directory on the filesystem.
 
         Parameters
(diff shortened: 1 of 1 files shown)


sccolbert:
See this comment: https://github.com/nucleic/enaml/issues/204#issuecomment-217332807

Subclassing is not needed here. If `kwargs` solves your problem, I don't see much reason to add this.

bburan:
Thanks. I did see that and preserved that functionality. The reason why I suggested this modification was to allow for a reusable file dialog with preset values (e.g., a preset "current directory", wildcard filter, etc.). This facilitates DRY by allowing for:

    class ImportFileDialog(FileDialogEx):
        name_filters = ['Some file (.ext)']

    def save_file(event):
        file_name = ImportFileDialog.get_save_file_name()

    def open_file(event):
        file_name = ImportFileDialog.get_open_file_name()

It also makes it easier to subclass the dialog if one wants to ensure that `get_save_file_name` returns a filename with the appropriate extension if the user didn't add it themselves.

sccolbert:
Just pass the keywords in your reusable function. A subclass is overkill
for this.

Further, assuming a constructor signature for a subclass is an anti
pattern. If someone subclasses the dialog, but requires positional
constructor args, the classmethods will break.

PR #208, opened 2016-05-24
Python 3 : C layer migration

MatthieuDartiailh:
This PR aims at making the C layer of enaml work on both Python 2 and 3. It is similar in shape to https://github.com/nucleic/atom/pull/65.

I suspect that something may be slightly wrong as I have seen in Python 3 cases in which a failed setattr (due to an exception in some handlers) was not reporting any exception (SystemError) while its Python 2 counterpart was. However I was not able to make a minimal example yet (so far they all work)

I largely edited the history to make it easier to review so I would like to thank @peterazmanov for its contribution (as it disappeared from the history)

### DIFF
--- a/enaml/src/alias.cpp
+++ b/enaml/src/alias.cpp
@@ -7,7 +7,15 @@
 |----------------------------------------------------------------------------*/
 #include <sstream>
 #include "pythonhelpersex.h"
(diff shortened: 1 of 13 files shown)


MatthieuDartiailh:
Closing superseded by #227

PR #209, opened 2016-05-24
Python 3 : python level

MatthieuDartiailh:
This PR aims at making the Python code of enaml cross-compatible. 

It uses the future package to handle things like metaclasses, string/bytes, etc. It also use the byteplay version for Python 3 updated by @serprex and @peterazmanov. Note that it does not touch the parser.

PS : I have not yet tested this branch, and I am not sure it is fully viable as is.

Once again many thanks to @peterazmanov

Based on #208


serprex:
Be warned that Python 3.6 is having a rather drastic change to bytecode layout: http://bugs.python.org/issue26647

I quickly threw together a byteplay wordcode port: https://github.com/serprex/byteplay/blob/master/wbyteplay.py but https://github.com/haypo/bytecode I feel should be used over byteplay see https://bytecode.readthedocs.io/en/latest/byteplay_codetransformer.html

MatthieuDartiailh:
I have been following those at python-dev. If I can get bytecode to work on python 2 I will probably do the change. But for the time being just getting 3.4 and 3.5 would be a huge step forward.

MatthieuDartiailh:
Closing superseded by #227

PR #210, opened 2016-05-24
Python 3 : parser updates

MatthieuDartiailh:
This is the BIG PR : 
- refactor of the lexer/parser to be both class based and use subclassing to avoid duplication in the parser code. With this whole syntax of Python 3 is supported is supported in enaml files, including function annotations and async/await.
- inline the bytecode for list/dict/set comprehension. This is because under Python those all use function calls. These calls cannot be passed the current dynamic scope, which makes comprehensions quite useless. Note that previously this was also true of dict/set comprehension on Python 2.

Based on #208 and #209


MatthieuDartiailh:
I fixed some issues on Python 3 in the tests and fixed the handling of comprehensions. Now the function is called with the enclosing DynamicScope as locals (the function code is made to have newlocals be false) As any intermediate value used in the comprehension is only stored in the fast locals there is no leak and no need to clone the scope. This is way cleaner than the inlining and allow things such as nested comprehension to work.
```python
[i for i in [j for j in range(10)]]
```

bburan:
If this is the same as the version on the ecpy conda channel, then I'm happy to report that I was able to port my application to Python 3 using this update with no issues.

MatthieuDartiailh:
Actually the version on the ecpy channel is a bit more recent as it features partial python 3.6 support (f-strings are missing) and qt5 support. But other than that there are identical.

bburan:
Ok. What can I do to help get this merged? For example, is there any testing that needs to be done? My application uses the Enaml workbench architecture, templates, widgets, etc., so I'm sure problems would have popped up by now if there were issues. I'm also using Python 3.6.

MatthieuDartiailh:
I wish I knew. @sccolbert is not answering lately.

MatthieuDartiailh:
ping @sccolbert

MatthieuDartiailh:
Closing superseded by #227

PR #213, opened 2016-05-25
Fix missing support for f(a, *b, c=1)

### LINKED ISSUE #212: Parser : function call f(a, b, *c, d=1) not supported
That form of function call is supported even under python 2. I think it is a matter of adding the right couple of rule to the parser. I can do it this week end if nobody beats me to it.

MatthieuDartiailh:
Close #212


enaml/core/parser.py
```diff
@@ -3444,6 +3444,92 @@ def p_arglist14(p):
         kwnames.add(kw.arg)
     p[0] = Arguments(keywords=keywords, starargs=p[2], kwargs=p[8])
 
+    def p_arglist15(self, p):
```
sccolbert: duplicate definition - looks like this function should be removed.

MatthieuDartiailh: Fixed sorry, left over pasting from fixing also #210

sccolbert:
The `arglist_list` needs to be validated using the `_validate_arglist_list` function (like the other rules), and I *think* we may also need to ensure there are no keyword args before a `*arg`, but I'm not sure on that. Would have to test that part on the real interpreter.

MatthieuDartiailh:
Indeed I will add the validation tomorrow. And check on the interpreter.

MatthieuDartiailh:
Added proper validation of the argument list  and also regenerated the parse_tab file. Hope all is well now, I should not have rushed yesterday and got some sleep first. Note that the fix is also present in #210

MatthieuDartiailh:
ping @sccolbert I think all comments are addressed. I actually checked in the interpreter and f(a, b=1, *c, d=2) is correct so there is no need for extras checks.

enaml/core/parser.py
```diff
@@ -3445,6 +3445,56 @@ def p_arglist14(p):
...
+def p_arglist17(p):
+    ''' arglist : arglist_list STAR test COMMA argument COMMA DOUBLESTAR test '''
+    args, kwargs = _validate_arglist_and_kwlist(p, p[1], [p[5]])
+    p[0] = Arguments(args=args, keywords=kwargs, starargs=p[3], kwargs=p[8])
+
+
+def p_arglist18(p):
+    ''' arglist : arglist_list STAR test COMMA arglist_list argument COMMA DOUBLESTAR test '''
+    args, kwargs = _validate_arglist_and_kwlist(p, p[1], p[5] + [p[6]])
+    p[0] = Arguments(args=args, keywords=kwargs, starargs=p[3], kwargs=p[8])
```
sccolbert: `p[8]` should be `p[9]` if I'm counting correctly

enaml/core/parser.py
```diff
@@ -3445,6 +3445,56 @@ def p_arglist14(p):
...
+    for arg in items:
+        if isinstance(arg, ast.keyword):
+            kws.append(arg)
+            kwnames.add(arg.arg)
+        else:
+            args.append(arg)
+    for kw in keywords:
+        if not isinstance(kw, ast.keyword):
+            msg = 'only named arguments may follow *expression'
+            tok = FakeToken(p.lexer.lexer, p.lineno(1))
```
sccolbert: you might need to pass the `lineno` as an argument to this function. Only literal tokens like `STAR`, `COMMA`, etc. will correctly report line numbers. In all the rules where this is invoked, `p.lineno(1)` will  be an expression token, which will not report the correct line number. This is limitation in ply.

MatthieuDartiailh: In all the considered cases STAR is in second position so using p.lineno(2) would work, no ?

enaml/core/parser.py
MatthieuDartiailh: Yes indeed, sorry about that.

enaml/core/parser.py
sccolbert: It would for now, but could lead to a bug in the future if we refactor. I *think* that's why inlined the keyword checks originally.

enaml/core/parser.py
```diff
@@ -3445,6 +3445,56 @@ def p_arglist14(p):
     p[0] = Arguments(keywords=keywords, starargs=p[2], kwargs=p[8])
 
 
+def _validate_arglist_and_kwlist(p, items, keywords):
```
sccolbert: Would you mind adding a comment about the assumption for the line number? That way we can update if we ever refactor this code.

sccolbert:
One last comment, then I'll merge.

MatthieuDartiailh:
Done. Thanks for your time.

PR #220, opened 2017-02-27
Fixed a typo in Looper example.

EmanueleCagliero:
The example printed the __repr__ of the (internal) fields of the Looper object. Instead now it prints the strings inserted in the GUI in the 'Items' field, as it was intended to do.


sccolbert:
The intent was to actually print the widgets generated by the Looper as a way to demonstrate the APIs.

PR #227, opened 2017-09-03
Add Python 3.4, 3.5, and 3.6 support 

MatthieuDartiailh:
This work is huge, I am not sure anybody can review it in full, but anybody willing is welcome. I did my best to clean things up but this is a rework of stuff I have been working on for several months. I will do a review myself and ping when the one week delay before merge will start.

This is based on #225

PR #228, opened 2017-09-03
Add Qt5 support

MatthieuDartiailh:
This adds qt5 support through the use of qtpy as a compatibility layer. As it is based on #227, the diff is huge I would recommend waiting for #227 to get in before reviewing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,9 @@ before_install:
 
   # Install enaml dependencies and include special widgets dependencies
   # We do not test vtk as it segfaults on Travis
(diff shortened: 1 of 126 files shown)


MatthieuDartiailh:
I need to update the selection of pytest-qt QT_API based on the one used by enaml for the tests to pass.

MatthieuDartiailh:
Rebased on top of master. Will merge in two weeks save if I get some comments/complaints.

MatthieuDartiailh:
I just re-tested everything on ecpy and found one minor mistake from the previous PR. I will merge and release.

PR #230, opened 2017-11-14
Fix live_editor.enaml example for Qt5

frmdstryr:


sccolbert:
Not sure this is the right fix for this.

MatthieuDartiailh:
I am not a big fan of fixing the issue in this way. Also could you point to a way to get QScintilla packaged with PyQt5 such that I add this to the build matrix.

frmdstryr:
Redid it in a proper way.  Also updated the test and travis.

I'm not sure why travis failed on 3.6, the errors seem unrelated.

sccolbert:
Looks good to me. I'll let Matthieu make the call on the build script.

MatthieuDartiailh:
Looks good to me. I will squash and merge and update the release notes. I will also increase the tests timeout to avoid the random failures on Travis. 
I will wait a month before cutting a new bugix release to see if people find other bugs in the new version.

PR #231, opened 2017-11-15
travis: test the version of qt5 available on conda

MatthieuDartiailh:
This was not the case anymore because QScintilla pulls pyqt 5.9  from PyPI.


MatthieuDartiailh:
This works as expected merging.

Issue #115, opened 2013-12-18
with enaml.imports() support executing from .zip file (PEP 441)


jason-s:
I typically package up my python projects as .zip files so it's easy for other people to execute them. (See http://www.python.org/dev/peps/pep-0441/ )

It appears as though the "with enaml.imports()" feature does not support this at present. It would be a useful thing to add.

MatthieuDartiailh:
Closed by #232

PR #232, opened 2017-11-15
Zipimports

frmdstryr:
I'm not sure why but git made me commit my other changes so this also includes adding autocomplete to the scintilla widget.

I can break it apart if needed.

### DIFF
--- a/enaml/core/import_hooks.py
+++ b/enaml/core/import_hooks.py
@@ -13,6 +13,8 @@ import os
 import struct
 import sys
 import types
+from zipfile import ZipFile
+
 
 from future.utils import with_metaclass, exec_
 
@@ -240,7 +242,7 @@ class EnamlImporter(AbstractEnamlImporter):
                 enaml_path = os.path.join(stem, leaf)
                 file_info = make_file_info(enaml_path)
                 if (os.path.exists(file_info.src_path) or
-                    os.path.exists(file_info.cache_path)):
+                        os.path.exists(file_info.cache_path)):
                     return cls(file_info)
 
         # We're trying a load a package
@@ -254,7 +256,7 @@ class EnamlImporter(AbstractEnamlImporter):
                 enaml_path = os.path.join(stem, leaf)
                 file_info = make_file_info(enaml_path)
                 if (os.path.exists(file_info.src_path) or
-                    os.path.exists(file_info.cache_path)):
+                        os.path.exists(file_info.cache_path)):
                     return cls(file_info)
 
     def __init__(self, file_info):
@@ -337,7 +339,27 @@ class EnamlImporter(AbstractEnamlImporter):
             timestamp = struct.unpack('i', cache_file.read(4))[0]
         return (magic, timestamp)
 
-    def compile_code(self):
+    def read_source(self, src_path):
+        """ Read source code from the given path 
+        
+        Parameters
+        ----------
+        src_path: string
+            The path or uri of the source file to be read by the importer
+        
+        Returns
+        -------
+        result : string or bytes
+            The source code to be passed to the parser.
+        
+        """
+        return read_source(src_path)
+
+    def get_src_mod_time(self, src_path):
+        """ Get the last modified time of the given source path """
+        return int(os.path.getmtime(src_path))
+
+    def compile_code(self, src_path):
         """ Compile the code object for the Enaml module and
         the full path to the module for use as the __file__ attribute
         of the module.
@@ -350,11 +372,44 @@ class EnamlImporter(AbstractEnamlImporter):
 
         """
         file_info = self.file_info
-        src_mod_time = int(os.path.getmtime(file_info.src_path))
-        ast = parse(read_source(file_info.src_path))
-        code = EnamlCompiler.compile(ast, file_info.src_path)
+        src_mod_time = self.get_src_mod_time(src_path)
+        ast = parse(self.read_source(src_path))
+        code = EnamlCompiler.compile(ast, src_path)
         self._write_cache(code, src_mod_time, file_info)
-        return (code, file_info.src_path)
+        return (code, src_path)
+
+    def get_cached_code(self, src_path):
+        """ Loads and returns the code object for the Enaml module from
+        the cache if it exists.
+        Paramters
+        ---------
+        src_path : string
+            Path where the source should exist if source code is present 
+        
+         Returns
+         -------
+         result : (code, path) or None
+             The Python code object for the .enaml module, and the full
+             path to the module as a string.
+        
+         """
+        # If the .enaml file does not exists, just use the .enamlc file.
+        # We can presume that the latter exists because it was already
+        # checked by the loader. Should the situation ever arise that
+        # it was deleted between then and now, an IOError is more
+        # informative than an ImportError.
+        file_info = self.file_info
+        if not os.path.exists(src_path):
+            code = self._load_cache(file_info)
+            return (code, src_path)
+
+        # Use the cached file if it exists and is current
+        src_mod_time = self.get_src_mod_time(src_path)
+        if os.path.exists(file_info.cache_path):
+            magic, ts = self._get_magic_info(file_info)
+            if magic == MAGIC and src_mod_time <= ts:
+                code = self._load_cache(file_info, set_src=True)
+                return (code, src_path)
 
     def get_code(self):
         """ Loads and returns the code object for the Enaml module and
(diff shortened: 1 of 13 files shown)


enaml/core/import_hooks.py
```diff
@@ -337,7 +339,27 @@ def _get_magic_info(self, file_info):
             timestamp = struct.unpack('i', cache_file.read(4))[0]
         return (magic, timestamp)
 
-    def compile_code(self):
+    def read_source(self, src_path):
```
MatthieuDartiailh: I do not see the point of passing explicitly the src_path, everything is in the file_info so I would simply use that.

enaml/core/import_hooks.py
```diff
@@ -337,7 +339,27 @@ def _get_magic_info(self, file_info):
...
+        
+        Returns
+        -------
+        result : string or bytes
+            The source code to be passed to the parser.
+        
+        """
+        return read_source(src_path)
+
+    def get_src_mod_time(self, src_path):
```
MatthieuDartiailh: Same here

enaml/core/import_hooks.py
```diff
@@ -337,7 +339,27 @@ def _get_magic_info(self, file_info):
...
+        """ Get the last modified time of the given source path """
+        return int(os.path.getmtime(src_path))
+
+    def compile_code(self, src_path):
```
MatthieuDartiailh: And here

enaml/core/import_hooks.py
```diff
@@ -350,11 +372,44 @@ def compile_code(self):
...
-        ast = parse(read_source(file_info.src_path))
-        code = EnamlCompiler.compile(ast, file_info.src_path)
+        src_mod_time = self.get_src_mod_time(src_path)
+        ast = parse(self.read_source(src_path))
+        code = EnamlCompiler.compile(ast, src_path)
         self._write_cache(code, src_mod_time, file_info)
-        return (code, file_info.src_path)
+        return (code, src_path)
+
+    def get_cached_code(self, src_path):
```
MatthieuDartiailh: Same here

enaml/core/import_hooks.py
```diff
@@ -350,11 +372,44 @@ def compile_code(self):
...
+        """ Loads and returns the code object for the Enaml module from
+        the cache if it exists.
+        Paramters
```
MatthieuDartiailh: Missing blank line before parameters section

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+    supported_archives = {
+        '.zip': ZipFile,
+    }
+
+    @classmethod
+    def locate_module(cls, fullname, path=None):
+        """ Searches for the given Enaml module within a zip and returns an instance
+        of this class on success.  
+        
+        Caching
```
MatthieuDartiailh: Caching is not recognized section in Numpydoc and will break the docs. You can simply omit the section.

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+                file_info = make_file_info(enaml_path)
+
+                # Strip package off path to get the archive name
+                archive_path = stem
+                for p in pkgpath:
+                    archive_path = os.path.dirname(archive_path)
+
+                # To check if cache file is in zip file
+                cache_path = os.path.relpath(file_info.cache_path,
+                                             archive_path).replace("\\","/")
```
MatthieuDartiailh: Missing space after comma

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+
+        # We're doing a direct import
+        else:
+            leaf = fullname + os.path.extsep + 'enaml'
+            for stem in sys.path:
+                enaml_path = os.path.join(stem, leaf)
+                file_info = make_file_info(enaml_path)
+                # To check if cache file is in zip file
+                cache_path = os.path.relpath(file_info.cache_path,
+                                             stem).replace("\\","/")
```
MatthieuDartiailh: Missing space after comma

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+                            name_list = archive.namelist()
+                            if ((leaf in name_list) or
+                                    (cache_path in name_list)):
+                                return cls(file_info,stem)
+                    except IOError:
+                        return
+
+    @classmethod
+    def _is_supported(cls, archive_path):
+        """ Checks if the given archive path is of one of the supported
```
MatthieuDartiailh: The dosctring content is not properly indented

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+            archive types. 
+            
+            Paramters
```
MatthieuDartiailh: Parameters (missing e)

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+            ---------
+            archive_path : string
+                The fully path to the archive.
+
+        Returns
+        -------
+        results : bool
+                Whether the archive is supported or not.
```
MatthieuDartiailh: Weird indent

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+
+        """
+        super(EnamlZipImporter, self).__init__(file_info)
+        self.archive_path = archive_path
+        self.archive = None  # Reference to opened archive
+
+    def get_src_mod_time(self, src_path):
+        """ Overridden to read the modified time of the archive 
+        instead of the source file.
+        """
```
MatthieuDartiailh: Missing blank line at the end of the docstring

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+        return super(EnamlZipImporter, self).get_src_mod_time(self.archive_path)
+
+    def read_source(self, src_path):
+        """ Overridden to read the source from the currently opened archive
```
MatthieuDartiailh: Same here

tests/test_zipimporter.py
```diff
@@ -0,0 +1,141 @@
...
+        for src, dst in [
+                ('examples/widgets/splitter.enaml',
+                    make_file_info('splitter.enaml').cache_path),
+                ('examples/widgets/notebook.enaml',
+                    make_file_info('package/subpackage/notebook.enaml').cache_path)
+            ]:
+            zf.writestr(dst, generate_cache(src), compress_type=zipfile.ZIP_DEFLATED)
+
+
+def with_library(f):
```
MatthieuDartiailh: This could be turned into a pytest fixture. I can do it if you are not familiar with them.

enaml/core/import_hooks.py
MatthieuDartiailh: Also why do you convert back slashes ? From the docs (that I skimmed quickly) ZipFile can handle windows path just fine.

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+
+                # To check if cache file is in zip file
+                cache_path = os.path.relpath(file_info.cache_path,
+                                             archive_path).replace("\\","/")
+
+                if (cls._is_supported(archive_path) and
+                        os.path.exists(archive_path)):
+
+                    # Path where code should be within the archive
+                    code_path = '/'.join(pkgpath+[leaf])
```
MatthieuDartiailh: Why not os.path.join ?

enaml/core/import_hooks.py
MatthieuDartiailh: In addition to the ```archive_path```, I would store this and the ```code_path```, on the importer to further reduce code duplication and avoid the need to pass a src_path around.

enaml/core/import_hooks.py
MatthieuDartiailh: See further comments for how to avoid code duplication in the ZipImporter

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+
+        Returns
+        -------
+        result : (code, path)
+            The Python code object for the .enaml module, and the full
+            path to the module as a string.
+
+        """
+
+        # Try to load from the cache on disk
```
MatthieuDartiailh: I am not sure to follow in which case this can actually apply. Could you explain it to me ?

enaml/core/import_hooks.py
```diff
@@ -368,26 +423,203 @@ def get_code(self):
...
+                code = marshal.loads(cache[8:])
+                return (code, code_cache_path)
+
+            #: Save reference
+            self.archive = archive
+
+            # Otherwise, compile from source and attempt to cache it on the system
+            code_path = os.path.relpath(file_info.src_path,
+                                        self.archive_path).replace("\\", "/")
+            result = self.compile_code(code_path)
```
MatthieuDartiailh: Where will the cache be written here ? To me it appears, this should lead to write the cache in the archive.

tests/test_zipimporter.py
frmdstryr: Yes that would be good.

frmdstryr:
Ok thanks for looking. I pulled out the scintilla changes and tried to address all the pep8 and doc issues. I also got rid of passing the `src_file` changes  as suggested and instead used an attribute and some method overrides.

The caching is a bit confusing. There's two possibilities:
1. `__enamlcache__` exists on the filesystem from previously generated cache when the code is compiled. I tried to retain this as it prevents having to open and read the zip file.
2. ` __enamlcache__`  exists within the `.zip` file. 

I'm not sure at the moment if the standard importer will cover importing from 1 even if the code was compiled from source within the zip file.  If this is the case, then the `get_cached_code` can be removed and left unchanged as the standard importer will find and use this. 

However the `source_exists` and `get_source_modified_time` methods are still required because these have to be done on the `.zip` archive and not where python expects the file to normally be on the file system.

As for the `.replace("\\","/")`. I wrote this a year ago so I don't remember exactly, I think this was because even though the ZipFile supports `\` separators it still had `/` separators in the zip file (but it may have been that [cx_Freeze](https://github.com/anthony-tuininga/cx_Freeze) was doing this). Nevertheless using replace ensures it always works (the app I used this for supported win, mac, & linux).

MatthieuDartiailh:
I guess what I find the more confusing about the first possibility you evoke concerning the cache is the location of the cache file. Does it live outside the archive ? and how did it end up there, and how can it be found ? 
Sorry for the perhaps stupid question but I fail to see how a cache file for a file inside an archive could be generated outside the archive and still be findable by the import machinery.
I will look over the changes tomorrow.

frmdstryr:
My understanding is the `EnamlImporter` attempts to create this in in https://github.com/nucleic/enaml/blob/master/enaml/core/import_hooks.py#L294.

frmdstryr:
I looked into this more, it turns out it always attempts to write the cache but fails since the `cache_file` and `cache_dir` point to within the `library.zip` and hence the situation in 1 never occurs (mkdir fails).

```
EnamlFileInfo(src_path='library.zip/buttons.enaml', cache_path='library.zip/__enamlcache__/buttons.enaml-py35-cv25.enamlc', cache_dir='library.zip/__enamlcache__')

EnamlFileInfo(src_path='library.zip/package/subpackage/slider.enaml', cache_path='library.zip/package/subpackage/__enamlcache__/slider.enaml-py35-cv25.enamlc', cache_dir='library.zip/package/subpackage/__enamlcache__')
```

So, should I make it write cache to outside the zip or is that undesired behavior?

MatthieuDartiailh:
That is what I suspected (and also why I failed to understand why we were looking for a cache outside the archive). I am not being familiar with working with zip archive but I guess it would be pretty surprising to modify the content of the archive. So I would say that we can bypass the first cache check and expect to fail writing the cache. However I guess we should first check how Python handles this case.

sccolbert:
I would follow whatever Python is doing. The `.enamlc` files and correlating logic basically mirrors python's `.pyc` logic. If Python's zip importer is not writing cache files, neither should we.

MatthieuDartiailh:
Looking at the Python ZipImporter https://github.com/python/cpython/blob/master/Modules/zipimport.c I do not seed any logic related to writing cache (but I may be wrong).

MatthieuDartiailh:
This historical PEP goes into the same direction.

MatthieuDartiailh:
I fixed some issues while porting the tests to pytest. 
First on windows all tests were failing because files were not opened with universal newline support. Second the importer was not honoring the encoding of the files. I don't yet have tests for this but I plan to add some at one point. I also fixed some PEP 8 issues.

frmdstryr:
Thanks, merged your changes and also removed the unnecessary caching checks and disabled the `_write_cache` since it doesn't work with the zip archive.

MatthieuDartiailh:
This is good to go for me. The only thing that could be improved is adding tests to check that we handle properly encoding declarations. @frmdstryr if you have time to do it feel free to do it.
I will merge in a week if nobody comment before.

MatthieuDartiailh:
@sccolbert any comment on this ?

sccolbert:
Looks generally fine to me. I didn't audit the logic, but if it works, 👍

MatthieuDartiailh:
Merging, thanks @frmdstryr for the contribution and sorry for the delay.

Closes #115

PR #233, opened 2017-11-16
Py3 menu sort

frmdstryr:
Fixes this crash
```
  File "/app/venv/lib/python3.5/site-packages/enaml/qt/q_deferred_caller.py", line 42, in customEvent
    event.callback(*event.args, **event.kwargs)
  File "/app/manifest.enaml", line 58, in start_default_workspace
    ui.select_workspace('micropyde.editor')
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/ui_plugin.py", line 157, in select_workspace
    new_workspace.start()
  File "/app/workspaces/editor/__init__.py", line 40, in start
    self.workbench.register(manifest)
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/workbench.py", line 61, in register
    self._add_extensions(manifest.extensions)
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/workbench.py", line 288, in _add_extensions
    self._update_extension_point(point, (), exts)
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/workbench.py", line 334, in _update_extension_point
    point.extensions = tuple(sorted(extensions, key=key))
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/ui_plugin.py", line 418, in _on_actions_updated
    self._refresh_actions()
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/ui_plugin.py", line 373, in _refresh_actions
    menus = create_menus(workbench, menu_items, action_items)
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/menu_helper.py", line 345, in create_menus
    return root.assemble()
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/menu_helper.py", line 272, in assemble
    return self.assemble_children()
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/menu_helper.py", line 238, in assemble_children
    for group, nodes in self.collect_child_groups():
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/menu_helper.py", line 202, in collect_child_groups
    yield group, solve_ordering(nodes)
  File "/app/venv/lib/python3.5/site-packages/enaml/workbench/ui/menu_helper.py", line 83, in solve_ordering
    flat.sort()
TypeError: unorderable types: MenuNode() < MenuNode()
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,9 @@ before_install:
      else
         $CONDA_INSTALL ipython=$IPY_VERSION qtconsole;
      fi'
-  - 'if [ $QT_VERSION > 4 ]; then
+
+  # Install QScintilla (for Qt5)
+  - 'if [ $QT_VERSION -gt 4 ]; then
          $PIP_INSTALL QScintilla;
      fi'
   - $PIP_INSTALL https://github.com/nucleic/atom/tarball/master
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ matrix:
   include:
     - env: PYTHON=2.7 QT_VERSION=4 IPY_VERSION=3
     - env: PYTHON=3.4 QT_VERSION=4 IPY_VERSION=3
-    - env: PYTHON=3.5 QT_VERSION=5 IPY_VERSION=5
-    - env: PYTHON=3.6 QT_VERSION=5 IPY_VERSION=6
+    - env: PYTHON=3.5 QT_VERSION=5 IPY_VERSION=5 QSCINTILLA=0
+    - env: PYTHON=3.6 QT_VERSION=5 IPY_VERSION=6 QSCINTILLA=1
 
 before_install:
 
@@ -34,6 +34,7 @@ before_install:
   # Install enaml dependencies and include special widgets dependencies
   # We do not test vtk as it segfaults on Travis
   - $CONDA_INSTALL pyqt=$QT_VERSION ply matplotlib
+
   # Install qtpy from pip to get a recent version (>= 1.3) on python 3.4
   - $PIP_INSTALL qtpy
   - 'if [ $IPY_VERSION < 4 ]; then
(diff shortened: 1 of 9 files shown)


sccolbert:
Why is this failing? Are generic objects no longer orderable based on id (pointer value) in Python 3?

sccolbert:
This change could lead to inconsistent layout when two different nodes have the same value. The existing implementation ensures, that for a given instance of a python process, the sort remains stable.

frmdstryr:
Not sure, but after putting this in I noticed the two menus switched around so I had to add another `after='file'` constraint, so this might not be the correct fix.  Unless the sort only applies when the solver has multiple options?

frmdstryr:
And yes it seems as if python 3 now doesn't allow sorting by id https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons

sccolbert:
ugh, what a senseless breakage.

The original version was minimizing the number of allocations during sorting. We can't do that anymore, so here's how I would change it (untested):

Replace:

```
     flat = []
    for node in nodes:
        node_var = variables[node.id]
        flat.append((node_var.value(), node))
    flat.sort(key=lambda n: n[0])

    return [pair[1] for pair in flat]
```

with

```
return sorted(nodes, key=lambda node: (variables[node.id].value(), id(node)))
```

frmdstryr:
Using `id` doesn't work, it's inconsistent. It displays in reverse sometimes.

sccolbert:
sorting by id alone is insufficient. tuples are sorted lexographically. the later elements are tie breakers for the earlier elements. in python 2, objects with out __lt__ or __cmp__ are sorted based on pointer value. python 3 breaks that, so we have to explicitly provide the id in python 3. we still need the variable value first.

frmdstryr:
Ok, your fix above seems to be working okay.

sccolbert:
Thanks!

frmdstryr:
Thank you.

PR #236, opened 2017-11-17
Improve dependencies constraints and make example less likely to randomly break

MatthieuDartiailh:


examples/widgets/popup_view.enaml
```diff
@@ -38,7 +38,7 @@ enamldef ConfigPopup(PopupView): popup:
     parent_anchor << POSITIONS[parent_box.selected_item]
     anchor << POSITIONS[view_box.selected_item]
     arrow_size << sizer.value
-    arrow_edge << arrow_edge.selected_item
+    arrow_edge << edge.selected_item
```
frmdstryr: Also tried this, doesn't seem to matter. Still fails randomly.

MatthieuDartiailh: Still with a recursion error ?

MatthieuDartiailh: I can reproduce but that becomes a name error.

sccolbert:
Can we reproduce the test failure on a local machine? I'm hesitant to make changes if something *only* fails on Travis. I get random Travis failures in my JS projects all the time, but the code is 100% correct. It's a Travis issue.

examples/widgets/popup_view.enaml
MatthieuDartiailh: I think it is a bug in sorted map since I can get this:
```python
class StandardTracedReadHandler(ReadHandler, HandlerMixin):
    """ An expression read handler which traces code execution.

    This handler is used in conjuction with the standard '<<' operator.

    """
    def __call__(self, owner, name):
        """ Evaluate and return the expression value.

        """
        func = self.func
        f_globals = func.__globals__
        f_builtins = f_globals['__builtins__']
        f_locals = self.get_locals(owner)
        tr = StandardTracer(owner, name)
        print(f_locals)
        print(self.scope_key, owner._d_storage,
              self.scope_key in owner._d_storage)
        scope = DynamicScope(owner, f_locals, f_globals, f_builtins, None, tr)
        return call_func(func, (tr,), {}, scope)
```
to print this which is obviously wrong:
```
<object object at 0x0000021D0D8FAEB0> sortedmap({<object object at 0x0000021D0D8FAEB0>: sortedmap({edge: <enaml.widgets.combo_box.ComboBox object at 0x0000021D0E928B88>, inc: <enaml.core.include.Include object at 0x0000021D0D8B6088>, offset_x: <enaml.widgets.spin_box.SpinBox object at 0x0000021D0E928048>, offset_y: <enaml.widgets.spin_box.SpinBox object at 0x0000021D0D85D8C8>, parent_box: <enaml.widgets.combo_box.ComboBox object at 0x0000021D0E928408>, popup: <enaml_test.ConfigPopup object at 0x0000021D0D83E2C8>, sizer: <enaml.widgets.slider.Slider object at 0x0000021D0E928EC8>, view_box: <enaml.widgets.combo_box.ComboBox object at 0x0000021D0E928FC8>}), _[anchor|trace]: <enaml.core.standard_tracer.SubscriptionObserver object at 0x0000021D0D88BB00>, _[arrow_size|trace]: <enaml.core.standard_tracer.SubscriptionObserver object at 0x0000021D0D88D0F0>, _[parent_anchor|trace]: <enaml.core.standard_tracer.SubscriptionObserver object at 0x0000021D0D88DBE0>}) False
```

MatthieuDartiailh:
It does fail locally and comes from sortedmap. Don't know how I missed that given that I believe I have been using this code for quite some time now.

examples/widgets/popup_view.enaml
frmdstryr: Correct it becomes a name error. I'm not sure how that ever worked? I always had to use a parent reference when doing an assignement when the two names are the same (`arrow_edge << arrow_edget.blabal`) like this:

```
enamldef ConfigPopup(PopupView): popup:
    foreground = 'white'
    background = 'rgba(30, 30, 30, 0.85)'
    parent_anchor << POSITIONS[parent_box.selected_item]
    anchor << POSITIONS[view_box.selected_item]
    arrow_size << sizer.value
    arrow_edge << popup.arrow_edge.selected_item
```
But that doesn't work either.

Seems like the scope is somehow not available.

MatthieuDartiailh:
Can it be related to the ordering issue we have seen previously ?

frmdstryr:
Yes also fails on my system (but intermittently), which is strange.

MatthieuDartiailh:
Same for me the failing appears to be random.

examples/widgets/popup_view.enaml
sccolbert: I can't really see what's going on in that output, could you format it a bit better?

sccolbert:
If it fails locally, then it's a bug. Can we whittle it down to a simple test case?

I wonder if it has to do with bytecode changes in Python3. I suspected a few of this sort to crop up.

examples/widgets/popup_view.enaml
MatthieuDartiailh: What appears wrong to me is that the owner is `<object object at 0x0000021D0D8FAEB0> ` which appears in `sortedmap({<object object at 0x0000021D0D8FAEB0>: sortedmap({edge: <enaml.widget ...` but the boolean test is False.

sccolbert: ah, yeah. it could be the comparison function inside sorted map is failing.

sccolbert: Hmm, seems that's already accounted for: 
https://github.com/nucleic/atom/blob/master/atom/src/pythonhelpers.h#L285-L309

MatthieuDartiailh: But sortedmap use some c++ comparison function on pointer no ?

sccolbert: ultimately, it uses Python's rich compare function from the C api. If that fails, it compares pointer values, which is the same as comparing `id(foo)` in Python.

MatthieuDartiailh: Yes but for bare object (like the scope key) less_equal makes no sense and comparing the ordering of pointer can return an unexpected answer. So sometimes we get the wrong pointer and the equality comparison fail which is why we get nothing.

MatthieuDartiailh: For the sake of testing replacing sortedmap by a usual dict appears to fix the issue. In addition, testing for the presence of the `self.scope_key` in the keys of the sortedmap in which it is not found work so somehow the sortedmap appear to be broken.

MatthieuDartiailh: I suspect that the root of the problem is that not all elements of the sorted map are not sorted based on the same ordering relation. For example in the failure above there are three keys in the map: one bare object and two strings. When ordering the map the comparison between the two strings use string ordering while any comparison to the base object will use the pointers. However the ordering of the strings based on pointers may differ from the ordering based on the Python comparison. As a consequence, the map can look unordered from the pointer point of view, which explain why the lookup can fail for the base object.
I do not have a good solution in mind for this. The issue will appear each time are inhomogeneous. At this point I wonder what the sortedmap buys us. If it is speed, we could always use pointers but I would like to see benchmarks in particular against 3.6 in which dict has been improved. If it has other uses we need a different fix.

sccolbert: In python 2, non-homogeneous lists are sortable, because objects can be totally ordered. It doesn't matter *how* the objects are compared, as long as they are consistently compared. What I bet is happening, is the on Python3 `PyObject_RichCompare` is failing when comparing an object and string, so I fall back on pointer compare. On Python 2, the rich compare function would compare based on type name (for incompatible types) before falling back on pointer address compare. 

Python 2:

capture


Python 3:

capture2

MatthieuDartiailh: How do you propose we fix the current bug then ?

sccolbert: Sorted map exists for memory efficiency. The memory footprint of an enaml app would grow substantially without it. We need to fix the comparison function we're using to be tolerant of this Python 3 change. I would start by seeing if Python 3 C-api exposes a function with the old behavior. Failing that, I would write a new comparison function that adds a check for type name.

MatthieuDartiailh: Python 3 does not expose any function with the old behavior. I opened https://github.com/nucleic/atom/pull/76 that fix the issue we are seeing here. It could use some review.

MatthieuDartiailh:
This is ready to be merged. I will merge in a week if nobody comments on it.

PR #239, opened 2017-11-23
Trace nested code objects

MatthieuDartiailh:
This allow to properly trace list/dict/set comprehensions which use a function. As a consequence the tracer has to be accessible from the function in the comprehension. 

Rather than modifying the function to of the comprehension to accept the tracer as argument, I chose to adapt the loading opcode based on whether or not we are dealing with a nested code object. To work, this requires the tracer to appear in the local namespace which requires a tiny trick to make it appear in the locals of the dynamicscope. This will have a penalty cost in term of speed for the traced comprehension but I believe this is acceptable.

The alternative would be to modify the function involved in the comprehension when the comprehension is involved in tracing, by changing its argument and modifying how it is called. It is feasible, but I am not sure it would be better than the solution I propose and I believe it would be more complex.

I will work on adding tests next.

### DIFF
--- a/enaml/core/code_tracing.py
+++ b/enaml/core/code_tracing.py
@@ -13,6 +13,7 @@ from .byteplay import (
 )
 
 
+from . import byteplay as bp
 if not IS_PY3:
     from .byteplay import DUP_TOPX
 else:
@@ -243,7 +244,7 @@ class CodeInverter(object):
         self.fail()
 
 
-def inject_tracing(codelist):
+def inject_tracing(codelist, nested=False):
     """ Inject tracing code into the given code list.
 
     This will inject the bytecode operations required to trace the
@@ -255,6 +256,9 @@ def inject_tracing(codelist):
     ----------
     codelist : list
         The list of byteplay code ops to modify.
+    nested : bool
+        Is the code modified defined inside an already traced code in which
+        case the tracer is in the local namespace and not the fast locals.
 
     Returns
     -------
@@ -262,6 +266,10 @@ def inject_tracing(codelist):
         A *new* list of code ops which implement the desired behavior.
 
     """
+    # If the code is nested into another already traced code, we need to use
+    # LOAD_NAME to access it rather than LOAD_FAST
+    tracer_op = LOAD_NAME if nested else LOAD_FAST
+
     # This builds a mapping of code idx to a list of ops, which are the
     # tracing bytecode instructions which will be inserted into the code
     # object being transformed. The ops assume that a tracer object is
@@ -274,7 +282,7 @@ def inject_tracing(codelist):
         if op == LOAD_ATTR:
             code = [                        # obj
                 (DUP_TOP, None),            # obj -> obj
-                (LOAD_FAST, '_[tracer]'),   # obj -> obj -> tracer
+                (tracer_op, '_[tracer]'),   # obj -> obj -> tracer
                 (LOAD_ATTR, 'load_attr'),   # obj -> obj -> tracefunc
                 (ROT_TWO, None),            # obj -> tracefunc -> obj
                 (LOAD_CONST, op_arg),       # obj -> tracefunc -> obj -> attr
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
I have added tests that I believe cover all cases. @frmdstryr do you see anything I may have missed ?

frmdstryr:
Looks good to me. The only thing I'd suggest is changing `test_list_comprehension_synchronization` to use eval to be consistent with the otheres

MatthieuDartiailh:
I don't use eval there because the order of the list is guaranteed which is not the case for set and dictionaries.

enaml/core/code_tracing.py
```diff
@@ -345,13 +353,31 @@ def inject_tracing(codelist):
...
             ]
             inserts[idx] = code
+        elif isinstance(op_arg, bp.Code):
+            # Inject tracing in nested code object.
+            # This handles the case of list/dict/set comprehensions that
+            # defines a nested function.
+
+            # As we cannot pass the tracer as argument to the function, get it
+            # from the dynamicscope that will find it in the fast locals and
+            # cache it. This will allow to access to it later
```
sccolbert: I  don't believe that DynamicScope caches lookups. If the code below `locals()` is returning the dynamic scope, you can instead just do `(LOAD_NAME, '__scope__')` which is special-cased by the dynamic scope to return itself. That said, I'm not sure where the dynamic scope is ever given a handle to a tracer in a way where it is "on the lookup path".

I'm also not sure that the nested functions are going to be called with the same `f_locals` object (which is the dynamic scope) without special work to set up the call.

sccolbert: Ah. It's the actual call to `locals()` which does the fetching of fast locals and storing them in slow locals. The internal call which does the work is `PyFrame_FastToLocalsWithError`.

So this approach seems to "technically work", except for understand how the dynamic scope gets passed to the nested function as the `f_locals`.

MatthieuDartiailh: This is handled as part of the support for comprehension. See compiler_utils for this part.

sccolbert: Ok. I'm still not a fan of how this is accomplished. I would still either rewrite the inner comprehension code to take `_[tracer]` as part of fast locals, *or*, update `DynamicScope` to make `_[tracer]` available as a "magic". Then you won't need this `locals()` dance.

MatthieuDartiailh: Do you have a preference between the two options (just to know before starting working on this again) ? I think the fix on the dynamic scope will be easier and probably look better.

sccolbert: The dynamic scope will definitely be the easier fix. But the `fast_locals` approach would make things more "uniform". I'll let you make the call.

MatthieuDartiailh:
I chose to go with the simple solution of simply making the tracer available in the dynamic scope as otherwise the changes to the bytecode would be a bit obscure and very dependent on the way the comprehension is handled (we would need first to alter the argument of the function by modifying the code object, identify the function call site, load the tracer and change the argument tuple building). It is possible but I feel it would be more complicated to follow than this solution.
@sccolbert  feel free to merge if you are happy with the changes. If you have some more minutes, you can also merge https://github.com/nucleic/atom/pull/76 if the changes agree with you.

enaml/src/dynamicscope.cpp
```diff
@@ -634,6 +634,10 @@ DynamicScope_getitem( DynamicScope* self, PyObject* key )
     if( strcmp( (char *)Py23Str_AS_STRING( key ), "__scope__" ) == 0 )
         return newref( pyobject_cast( self ) );
 
+    // _[tracer] magic
+    if( strcmp( (char *)Py23Str_AS_STRING( key ), "_[tracer]" ) == 0 )
```
sccolbert: It's possible for `self->tracer` to be `null` so I think you want:

`if( self->tracer && strcmp(...) == 0)`

enaml/src/dynamicscope.cpp
```diff
@@ -720,6 +724,10 @@ DynamicScope_contains( DynamicScope* self, PyObject* key )
     if( strcmp( (char *)Py23Str_AS_STRING( key ), "__scope__" ) == 0 )
         return 1;
 
+    // _[tracer] magic
+    if( strcmp( (char *)Py23Str_AS_STRING( key ), "_[tracer]" ) == 0 )
```
sccolbert: ditto

enaml/src/dynamicscope.cpp
MatthieuDartiailh: Fixed

PR #243, opened 2017-11-27
Add themes, autocomplete, markers, indicators, and show_line_numbers to scintilla

frmdstryr:
Preview
image

Themes are parsed and generated from http://richleland.github.io/pygments-css/


examples/applib/live_editor.enaml
```diff
@@ -20,7 +20,7 @@ from enaml.layout.api import HSplitLayout, AreaLayout, DockBarLayout
 from enaml.widgets.api import Window, Container, DockArea, DockItem
 
 
-MODEL_TEXT = """\
+MODEL_TEXT = """
```
MatthieuDartiailh: Why this change ?

examples/applib/live_editor.enaml
```diff
@@ -30,7 +30,7 @@ class DemoModel(Atom):
 """
 
 
-VIEW_TEXT = """\
```
MatthieuDartiailh: Same here ?

enaml/scintilla/scintilla.py
```diff
@@ -115,11 +129,11 @@ class Scintilla(Control):
 
     #: The theme to apply to the widget. See the './THEMES' document
     #: for how to create a theme dict for the widget.
-    theme = d_(Typed(dict, ()))
+    theme = d_(Dict())
```
MatthieuDartiailh: The container used by atom for Dict is not  dict subclass which sometimes causes issues. Is this not why it was not used here before ?

enaml/scintilla/scintilla.py
```diff
@@ -104,6 +112,12 @@ class Scintilla(Control):
     on this widget. All styling is supplied via the 'theme' attribute.
 
     """
+    #: Enable autocompletion
+    autocomplete = d_(Enum('none', 'all', 'document', 'apis'))
+
+    #: Autocompletion values
+    autocompletions = d_(List(str))
```
MatthieuDartiailh: I wish there was a way to hook an autocompletion library (such as jedi) in Scintilla... but it does not appear possible.

enaml/qt/qt_scintilla.py
```diff
@@ -27,10 +27,10 @@
 else:
     import QScintilla as Qsci
 
+from .QtGui import QColor, QFont, QPixmap
 
-from .QtGui import QColor, QFont
-
-from .q_resource_helpers import QColor_from_Color, QFont_from_Font
+from .q_resource_helpers import QColor_from_Color, QFont_from_Font, \
```
MatthieuDartiailh: I would prefer: 
```python
from .q_resource_helpers import (QColor_from_Color, QFont_from_Font,
                                                      get_cached_qimage)
```

enaml/scintilla/scintilla.py
frmdstryr: Never heard of jedi until now. IPython also has a [completer](https://github.com/ipython/ipython/blob/master/IPython/core/completer.py) that I've used.  Looks like either can be used. I'll see if I can add a examples.

enaml/scintilla/scintilla.py
frmdstryr: Not sure why I changed that. Will undo it.

sccolbert:
It doesn't make sense to me that all themes are put in the `idle_theme` file. Why not make a separate file for each theme?

frmdstryr:
Good idea, maybe `themes` package with a module for each theme?

sccolbert:
That sounds fine.

frmdstryr:
Should I include the script used to generate the themes? 

Also I added the `cursor_position` as it's needed to any sort of useful autocompletion.

image

frmdstryr:
Still have to work out some kinks, but jedi is awsome:

image

frmdstryr:
Also added indicators and markers

[Preview](https://youtu.be/YE7fgf5pZ9I)

MatthieuDartiailh:
@frmdstryr is this ready for review ?

frmdstryr:
The only thing that I haven't added yet is handling the marker and indicator signals (see [QScintilla2 docs](http://pyqt.sourceforge.net/Docs/QScintilla2/classQsciScintilla.html)).  Any suggestions on a way to handle the `Qt::KeyboardModifiers`? Is it recommended to pass a dictionary with named arguments to pass event data like `Event(dict)`  or just copy the signature `Event(int, int, modifiers)`?

There's also [annotations](http://www.scintilla.org/ScintillaDoc.html#Annotations) but I didn't like how that worked at all (it just added a line of read only text below the current line and was not styled). I think enaml's popups are better to use so I wasn't planning to add annotations.
 
Finally, using the name `autocompletions` always makes PyCharm say there's a spelling error, any suggestions for a better name?

MatthieuDartiailh:
I will try to give this a look next week.

enaml/applib/live_editor_model.py
```diff
@@ -232,6 +238,38 @@ def refresh_view(self):
...
 
+    def autocomplete(self, source, position):
+        """ Obtain autocompletion suggestions for the source text 
+        using jedi if available.
+        
+        """
+        if not COMPLETION_AVAILABLE or not source:
+            return []
+        try:
+            #: Use jedi to get suggestions
```
MatthieuDartiailh: Please use `#:` for documenting attributes but no for general comments. The point in using `#:` for attribute is because Sphinx can pick those up when generating documentation.

enaml/applib/live_editor_model.py
```diff
@@ -232,6 +238,38 @@ def refresh_view(self):
...
+            line, column = position
+            script = jedi.Script(source, line+1, column)
```
MatthieuDartiailh: By curiosity, as I have not yet tested this, do this does not slow down too much the edition ?

enaml/applib/live_editor_view.enaml
```diff
@@ -47,22 +49,40 @@ enamldef EditorPanel(Container):
...
             "tab_width": 4,
             "use_tabs": False,
             "indent": 4,
             "tab_indents": True,
             "backspace_unindents": True,
+            "autocompletion_threshold": 1,
+            "show_line_numbers": True,
         }
+        autocomplete = 'all'
+        autocompletions = ['enamldef']
```
MatthieuDartiailh: I believe this should include all enaml keywords (attr, alias, enamldef, func, template, event). WDYT ?

enaml/qt/qt_scintilla.py
```diff
@@ -27,10 +27,10 @@
 else:
     import QScintilla as Qsci
 
-
 from .QtGui import QColor, QFont
 
-from .q_resource_helpers import QColor_from_Color, QFont_from_Font
+from .q_resource_helpers import (QColor_from_Color, QFont_from_Font,
+    get_cached_qimage)
```
MatthieuDartiailh: Indentation does not match PEP8 recommendations.

enaml/qt/qt_scintilla.py
```diff
@@ -360,6 +479,97 @@ def set_text(self, text):
...
+        
+        """
+        self.widget.setAutoCompletionSource(AUTOCOMPLETION_SOURCE[mode])
+
+    def set_autocompletions(self, options):
+        """ Set the autocompletion options for when the autocompletion mode
+        is in 'all' or 'apis'.
+            
+        """
+
```
MatthieuDartiailh: remove blank line

enaml/scintilla/scintilla.py
```diff
@@ -8,11 +8,13 @@
...
 from atom.api import (
-    Atom, Int, Constant, Enum, Event, Typed, ForwardTyped, observe, set_default
+    Atom, Int, Constant, Enum, Event, Typed, List, ForwardTyped, Tuple,
+    Unicode, observe, set_default
 )
-
+from enaml.image import Image
 from enaml.core.declarative import d_
 from enaml.widgets.control import Control, ProxyControl
+from future.builtins import str
```
MatthieuDartiailh: Please keep external (external to enaml) grouped before enaml imports it helps identifying the dependencies faster.

enaml/scintilla/scintilla.py
```diff
@@ -66,6 +68,39 @@ class ScintillaDocument(Atom):
     uuid = Constant(factory=lambda: uuid.uuid4().hex)
 
 
+class ScintillaIndicator(Atom):
+    """ An indicator descriptor.
+    
+    """
+
```
MatthieuDartiailh: blank line

enaml/qt/qt_scintilla.py
```diff
@@ -335,13 +437,30 @@ def pull_color(name, default):
...
             pull_enum(INDENTATION_GUIDES, 'indentation_guides', 'none'))
+        w.setAutoIndent(pull_bool('auto_indent', False))
 
         # White Space
         send(w.SCI_SETVIEWWS, pull_enum(WHITE_SPACE, 'view_ws', 'invisible'))
         send(w.SCI_SETWHITESPACESIZE, pull_int('white_space_size', 1))
         send(w.SCI_SETEXTRAASCENT, pull_int('extra_ascent', 0))
         send(w.SCI_SETEXTRADESCENT, pull_int('extra_descent', 0))
 
+        # Autocompletion
```
MatthieuDartiailh: The widget documentation points the user the the SETTINGS file for documentation about the settings, but you not appear to have updated this file, could you do it ?

enaml/applib/live_editor_model.py
frmdstryr: Thanks. I just got into the habit of doing it that way, will change it.

enaml/applib/live_editor_model.py
frmdstryr: I haven't tested it with any very large files but I haven't noticed any slowdowns. .

enaml/applib/live_editor_view.enaml
frmdstryr: The `autocompletions` get's updated by the model so this is really only there so you can see how it works.

MatthieuDartiailh: Ok in this case what bothers me is that jedi will not propose enaml keyword for completion and in the live editor it would be nice to have them.

frmdstryr: Yes, I'd also like this. 

The keywords could simply be added [here](https://github.com/frmdstryr/enaml/blob/latest/enaml/applib/live_editor_model.py#L254) which will make it suggest them all the time. The problem is that that will make them show up everywhere, when the results should be context aware. It would be even more powerful if it could intelligently autocomplete declarative members when you're in a childef block.  

I did look though jedi's source to see how involved this would be and found that Jedi abstracts this part out into it's https://github.com/davidhalter/parso library.  Maybe could be somehow tied in with the enaml parser but that's as far as I got into it.

MatthieuDartiailh:
Let me know when this is ready for a new round of review. This probably won't be in 0.10.2 that I hope to release either this week or next week but hopefully we can include it soon after.

frmdstryr:
I don't plan on changing more. I merged https://github.com/nucleic/enaml/pull/245 into this branch and then reverted it as it wasn't pulled in yet so the changes in https://github.com/nucleic/enaml/pull/243/commits/b2559001afcf048a7d1541bd88b8ac9837d97818 are the only changes from the previous time you reviewed it.

enaml/applib/live_editor_view.enaml
MatthieuDartiailh: I somehow though that Scintilla was responsible for filtering the list of possible auto completion it was provided with to make meaningful suggestions. Anyway I still need to take some time to test this myself before merging. Hopefully it will not be too long.

MatthieuDartiailh:
Sorry @frmdstryr I missed the fact you addressed all my comments. I will test manually and see if I find anything clever to do about the auto-completion, I will keep you posted.

MatthieuDartiailh:
I tested locally on the applib example and this is great. I just made a small modification to the autocompletion to always propose enaml keywords (scintilla filter out meaningless completion) and I added some missing keywords in the scintilla lexers. If you are happy with those changes I will merge during the week.

MatthieuDartiailh:
Thanks a lot for this @frmdstryr

PR #245, opened 2017-11-30
Fixes for closure in comprehensions and tracing of lambda functions

MatthieuDartiailh:
Wrap functions defined inside operators or declarative function to call them with their scope of definition. This allows to handle properly comprehensions and lambdas (whose body is now properly traced). Also ensure that we compile the body of the :: operator as a function to properly handle closure.

Bump the compiler version to 26

I added tests for all the new behaviors

### DIFF
--- a/enaml/core/compiler_common.py
+++ b/enaml/core/compiler_common.py
@@ -684,6 +684,7 @@ def gen_operator_binding(cg, node, index, name):
         code = compile(node.value.ast, cg.filename, mode=mode)
         for op, op_arg in bp.Code.from_code(code).code:
             if isinstance(op_arg, bp.Code):
+                rewrite_globals_access(op_arg, global_vars)
                 code = op_arg.to_code()
                 break
     else:
--- a/enaml/core/compiler_common.py
+++ b/enaml/core/compiler_common.py
@@ -12,7 +12,7 @@ import ast
 
 from atom.api import Str, Typed
 
-from ..compat import IS_PY3
+from ..compat import IS_PY3, USE_WORDCODE
 from . import byteplay as bp
 from .code_generator import CodeGenerator
 from .enaml_ast import (
@@ -56,13 +56,19 @@ COMPILE_MODE = {
 
 #: Ast nodes associated with comprehensions which uses a function call that we
 #: will have to call in proper scope
-_COMP_NODES = [ast.ListComp] if IS_PY3 else []
+_FUNC_DEF_NODES = [ast.Lambda]
+if IS_PY3:
+    _FUNC_DEF_NODES.append(ast.ListComp)
 if hasattr(ast, 'DictComp'):
-    _COMP_NODES.append(ast.DictComp)
+    _FUNC_DEF_NODES.append(ast.DictComp)
 if hasattr(ast, 'SetComp'):
-    _COMP_NODES.append(ast.SetComp)
+    _FUNC_DEF_NODES.append(ast.SetComp)
 
-_COMP_NODES = tuple(_COMP_NODES)
+_FUNC_DEF_NODES = tuple(_FUNC_DEF_NODES)
+
+#: Opcode used to create a function
+_MAKE_FUNC = ((bp.MAKE_FUNCTION,) +
+              ((bp.MAKE_CLOSURE,) if not USE_WORDCODE else ()))
 
 
 def unhandled_pragma(name, filename, lineno):
@@ -390,20 +396,23 @@ def safe_eval_ast(cg, node, name, lineno, local_names):
         cg.code_ops.extend(expr_cg.code_ops)
 
 
-def analyse_globals_and_comps(pyast):
-    """Collect the explicit 'global' variable names and check for the presence
-    of comprehensions (list, dict, set).
+def analyse_globals_and_func_defs(pyast):
+    """Collect the explicit 'global' variable names and check for function
+    definitions
+
+    Functions definition can exist if a comprehension (list, dict, set) is
+    present or if a lambda function exists.
 
     """
     global_vars = set()
-    has_comp = False
+    has_def = False
     for node in ast.walk(pyast):
         if isinstance(node, ast.Global):
             global_vars.update(node.names)
-        elif isinstance(node, _COMP_NODES):
-            has_comp = True
+        elif isinstance(node, _FUNC_DEF_NODES):
+            has_def = True
 
-    return global_vars, has_comp
+    return global_vars, has_def
 
 
 def rewrite_globals_access(code, global_vars):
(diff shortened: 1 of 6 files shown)


MatthieuDartiailh:
@sccolbert in theory this could allow to define function in declarative functions and event handler but I do not see a compelling use case to remove this restriction.

frmdstryr:
For callbacks to async functions it would be nice.

MatthieuDartiailh:
@frmdstryr could you test this ?

frmdstryr:
Tests are passing in 2.7 and 3.5 for me.

frmdstryr:
Also undid my previous workarounds for the listcomp and local var issue and both are working :+1:

MatthieuDartiailh:
@sccolbert this is the last pending PR that needs to be in before the next bugfix release that I plan mid-december. It would be terrific if you could give it a look.

enaml/core/compiler_helpers.py
```diff
@@ -755,5 +778,5 @@ def add_decl_function(node, func, is_override):
     'validate_spec': validate_spec,
     'validate_template': validate_template,
     'validate_unpack_size': validate_unpack_size,
-    'call_func': call_func,
```
sccolbert: there's no need to remove this, just add the new helper.

enaml/core/compiler_helpers.py
```diff
@@ -736,6 +739,26 @@ def add_decl_function(node, func, is_override):
...
+        Function for which a new scope need to be built
+
+    scope : DynamicScope
+        Scope in which the function was defined.
+
+    """
+    def wrapper(*args, **kwargs):
+        return call_func(func, args, kwargs, scope)
+
+    update_wrapper(wrapper, func)
```
sccolbert: why this over `@functools.wraps`?

enaml/core/parser/base_parser.py
```diff
@@ -779,13 +779,25 @@ def p_operator_expr2(self, p):
...
+        func_node.args = self._make_args([])
+        func_node.decorator_list = []
+        if IS_PY3:
+            func_node.returns = None
+        func_node.body = p[2]
+        func_node.lineno = lineno
+
+        mod = ast.Module(body=[func_node])
+        ast.fix_missing_locations(mod)
+
```
sccolbert: I don't think this is correct. Can you give an example of the closure variables you're hoping to pick up with this change?

Also, I don't think this is going to work as you expect. When the `::` operator is run to generate the expression handler, it wraps the generated bytecode in it's own function, which is then invoked by the handler: https://github.com/nucleic/enaml/blob/master/enaml/core/operators.py#L70  So each time the notification is triggered, this new internal function is going to be defined. I'm not entirely sure how this is passing tests.

sccolbert: But your example already works without this change, because all `LOAD_GLOBALS` are re-written to `LOAD_NAME`.

sccolbert: Can you give me an example of something that doesn't work currently, that this change fixes? Also, can you link me to where you "extract the body of the function and pass it forward"?

sccolbert: Ah, okay, you mean that doesn't work on Python 3. Sorry.

We just need to recursively rewrite LOAD_GLOBAL to LOAD_NAME for nested code objects.

MatthieuDartiailh: It is not so simple because LOAD_NAME fails to access fast locals of the outer scope. I will answer in more details later. I am attending a conference this week.

sccolbert: I don't think there's anything in fast locals that the inner code will need to access.

MatthieuDartiailh: Still looking at this kind of comprehension:
```python
a = 1
[a for i in range (2)]
```

Digging deeper I found that:
- when Python compile the code as if it was a top-level module (which it does without this fix), it uses STORE_GLOBAL and LOAD_GLOBAL to handle `a`
- to unify the handling of comprehensions between operators and declarative functions I refactored the handling of global variables such that as LOAD_GLOBAL and STORE_GLOBAL are replaced by LOAD_NAME/STORE_NAME save if the variable is marked using the `global` keyword, both in declarative functions and operators. Currently this is not done if there is no comprehension in an operator which is inconsistent. cf https://github.com/nucleic/enaml/blob/master/enaml/core/compiler_common.py#L685
- finally when creating the operator (https://github.com/nucleic/enaml/blob/master/enaml/core/operators.py#L156), we optimize the code for locals (https://github.com/nucleic/enaml/blob/master/enaml/core/operators.py#L49) which replaces LOAD_NAME/STORE_NAME for local variables by LOAD_FAST/STORE_FAST.

As a consequence a is not stored in the namespace, which explains the failure.

From my point of view the proposed fixed is the more correct since the body of a notification operator is truly a function body (even without return and yield). Alternatively we could suppress the global variable analysis in operators. 

I will update the code to always analyse global variables in operators for the time being, so that the current behavior is consistent.

MatthieuDartiailh:
I believe I have answered all your questions in my last two posts @sccolbert, could you see if that answers your concerns.

enaml/core/compiler_helpers.py
MatthieuDartiailh: I removed it because I was the one to add it in the first place when adding first support for comprehensions.

enaml/core/compiler_helpers.py
MatthieuDartiailh: No reason I will use `functools.wraps`

MatthieuDartiailh: Actually it is because wraps expects the decorator to take as single argument the function.

MatthieuDartiailh:
ping @sccolbert 
I guess you are fairly busy but do you think you could review this so that we get the bugfix release out before Christmas.

MatthieuDartiailh:
ping @sccolbert 
I would like to make the next release at the beginning of February and this PR is really needed.

enaml/core/compiler_common.py
```diff
@@ -679,13 +677,25 @@ def gen_operator_binding(cg, node, index, name):
...
-    global_vars, has_comp = analyse_globals_and_comps(node.value.ast)
-    code = compile(node.value.ast, cg.filename, mode=mode)
+    global_vars, has_defs = analyse_globals_and_func_defs(node.value.ast)
+    # In mode exec, the body of the operator has been wrapped in a function def
+    # after the compilation we extract the function code
+    if mode == 'exec':
+        code = compile(node.value.ast, cg.filename, mode=mode)
+        for op, op_arg in bp.Code.from_code(code).code:
+            if isinstance(op_arg, bp.Code):
+                rewrite_globals_access(op_arg, global_vars)
```
sccolbert: Is this duplicating logic from line 697 below?

enaml/core/compiler_helpers.py
sccolbert: :+1:

enaml/core/compiler_common.py
MatthieuDartiailh: Looks like it is indeed. I will fix it.

Issue #251, opened 2018-01-19
Python function on enaml file with multiple kwargs causes parser to crash.


gabrielcnr:
I've created a new environment with the latest version of `enaml` (`0.10.1`) pulled from `conda-forge` channel, using Python `2.7.14` on my Mac OS (I also verified the potential bug on Windows too).

I found out if I have a Python function defined on the `.enaml` module with 2 specified keyword arguments with defaults alongside the optional `**kwargs`, the parser fails with an `AttributeError`.

Please consider the following example (`example.enaml`):

```python
from enaml.widgets.api import *

def foobar(first=1, second=2, **kwargs):
    pass


enamldef Main(Window):
    title = 'Enaml bug'
```

When executing it using `enaml-run example.enaml` it should exit with the following error traceback:

```
Traceback (most recent call last):
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/bin/enaml-run", line 6, in <module>
    sys.exit(enaml.runner.main())
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/lib/python2.7/site-packages/enaml/runner.py", line 46, in main
    ast = parse(enaml_code, filename=enaml_file)
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/lib/python2.7/site-packages/enaml/core/parser/__init__.py", line 48, in parse
    return _parser.parse(enaml_source, filename)
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/lib/python2.7/site-packages/enaml/core/parser/base_parser.py", line 245, in parse
    return self.parser.parse(source, debug=0, lexer=lexer)
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/lib/python2.7/site-packages/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/lib/python2.7/site-packages/ply/yacc.py", line 1118, in parseopt_notrack
    p.callable(pslice)
  File "/Users/gabriel/anaconda/anaconda/envs/enaml_bug/lib/python2.7/site-packages/enaml/core/parser/base_parser.py", line 3356, in p_varargslist18
    p[0] = self._make_args(args, defaults=defaults, kwarg=p[7])
[...]

PR #252, opened 2018-01-20
Fixes a couple of errors in the parser handling of function definition

MatthieuDartiailh:
One mistake was reported in #251, I found the other while debugging. @frmdstryr feel free to give a look and if you have an idea on how we could test this in clean way, I am very much interested.


gabrielcnr:
@MatthieuDartiailh  I wrote a test for this which is probably closer to an integration test rather than a unit test because it tests the parser from a functional POV used with the compiler instead of making punctual assertions against AST types.

What do you think about the below:

```python
from enaml.core.parser import parse, write_tables
from enaml.core.enaml_compiler import EnamlCompiler
from future.utils import exec_


def test_p_varargslist18():
    write_tables()

    source = """
def func(foo=1, bar=2, **kwargs):
    d = {'foo': foo, 'bar': bar}
    d.update(kwargs)
    return d
"""

    ast = parse(source, filename='foo.enaml')
    code = EnamlCompiler.compile(ast, 'foo.enaml')

    ns = {}
    exec_(code, ns)

    func = ns['func']

    assert {'foo': 1, 'bar': 2} == func()

    assert {'foo': 100, 'bar': 200, 'spam': 300} == func(100, 200, spam=300)

```

MatthieuDartiailh:
Thanks for your interest @gabrielcnr. Actually reading your proposed test gave me an idea for something that is believe is more robust and easier to automatize (based on ast analysis). I found a bunch of other issues running the tests actually. If you could take the time to read through the tests to check that I have not forgotten anything it would be great.

frmdstryr:
Overall I'd say it looks good.   Is there a reason it's in a separate package? Took me a little to find it there...

The codecov report shows a few cases that are missing https://codecov.io/gh/nucleic/enaml/tree/e75c88e51ddd06daa0b31fde4dafeef503b22d1f/enaml/core/parser

frmdstryr:
I added a few more cases here: https://github.com/MatthieuDartiailh/enaml/compare/parser_fixes...frmdstryr:parser_fixes?expand=1 but it only increases coverage by about 20 lines. 

The last two statements don't pass the ast check so IDK what's going on there.

MatthieuDartiailh:
Thanks @frmdstryr. I put the tests in a different package because I usually try to follow the files organization for the tests organizations. So far we have little enough tests that it does not matter, but it is worth starting to clean up a bit.
I was also not planning to add total coverage of the parser in this PR (simply of the function definitions). If you are interested in adding more tests, feel free to open another pull request. I will merge this as is and look into the two cases that fail for you later.

PR #253, opened 2018-01-25
Get rid of a spurious print triggerred by setting super in DynamicScope

MatthieuDartiailh:
~~I identified the issue recently on Python 3.6. To reproduce run~~
```pytest tests\test_layout_container.py -s```

~~I must say I do not understand the fix.~~ @sccolbert ~~I could use your deep knowledge of the C API here~~. ~~This was due to some weird leftover in my working environment~~

Now using the branch to see if we can get the tests to pas on pyqt 5.10


MatthieuDartiailh:
Getting pyqt 5.10 won't be that easy apparently...

PR #254, opened 2018-02-24
ImageView: Add constraints to enforce the respect of aspect ratio when required

MatthieuDartiailh:
Previously if the image was made smaller than the image size, the size hint (which reflects the image size before rescaling) was used by the parent container which led to oversized container. This patch fixes this.

### DIFF
--- a/enaml/qt/qt_image_view.py
+++ b/enaml/qt/qt_image_view.py
@@ -263,3 +263,11 @@ class QtImageView(QtControl, ProxyImageView):
 
         """
         self.widget.setPreserveAspectRatio(preserve)
+
+    def get_aspect_ratio(self):
+        """Get the image aspect ratio from the pixmap.
+
+        """
+        pixmap = self.widget.pixmap()
+        pm_size = pixmap.size()
+        return pm_size.width()/pm_size.height()
--- a/enaml/widgets/image_view.py
+++ b/enaml/widgets/image_view.py
@@ -32,6 +32,9 @@ class ProxyImageView(ProxyControl):
     def set_preserve_aspect_ratio(self, preserve):
         raise NotImplementedError
 
+    def get_aspect_ratio(self):
+        raise NotImplementedError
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
In particular this now produce the expected result:

```python
from enaml.image import Image
from enaml.layout.api import hbox
from enaml.widgets.api import Window, Container, PushButton, ImageView
import struct

image_width = 800
image_height = 600
image_data = b''
for y in range(image_height):
    for x in range(image_width):
        image_data += struct.pack('BBBB', 255, x*255//image_width,
                              y*255//image_height, 255)

enamldef Main(Window):
    Container:
        constraints = [hbox(buttons, image),
                       (buttons.height == image.height) | 'required']
        Container: buttons:
            hug_height = 'required'
            PushButton:
                text = 'First'
            PushButton:
                text = 'Second'
            PushButton:
                text = 'Third'
        ImageView: image:
            scale_to_fit = True
            allow_upscaling = True
            preserve_aspect_ratio = True
            image = Image(format='argb32', data=image_data,
                          raw_size=(image_width, image_height))

```

frmdstryr:
Works here.  

On a side note the python 3 version takes a whopping 2 minutes to open (python 2 takes 3 seconds). 

```
(venv) jrm@xps:~/Workspace/enaml$ time enaml-run tests/example.enaml 
real    0m3.105s
user    0m0.781s
sys     0m0.128s

(venv3) jrm@xps:~/Workspace/enaml$ time enaml-run tests/example.enaml 
real    1m59.961s
user    0m46.260s
sys     1m10.602s

```
Is struct pack / appending bytes really that slow on py3?

MatthieuDartiailh:
I noticed it was super slow but it is the example posted by the OP on the mailing list. I did not take the time to look at it in more details. Thanks for testing @frmdstryr. I hope to make a final review and merge #243 this weekend.

PR #259, opened 2018-03-19
Automatically show MdiWindow when added to MdiArea (if visible)

### LINKED ISSUE #147: MDIWindows are not automatically shown when added to an MDIArea
This should work, but doesnt:
(adding a .show() after appending the window is workaround)

``` python
from enaml.widgets.api import (
    Window, Container, MdiArea, MdiWindow,
    PushButton, CheckBox, Field, Html, GroupBox
    )
from enaml.core.api import Include
from enaml.layout.api import vbox, hbox, align, spacer

enamldef MdiContent(MdiWindow):
    attr num: int
    Container:
        constraints = [
            vbox(
                hbox(pb, cb, fld),
                html,
            ),
            align('v_center', pb, cb, fld),
        ]
        PushButton: pb:
            text = 'Button'
        CheckBox: cb:
            text = 'Activate'
        Field: fld:
            pass
        Html: html:
            source = '<h1>New window %d</h1>'%num


enamldef Main(Window): main:
    attr count: int = 0
    Container:
        constraints = [
            vbox(hbox(add_btn, spacer),
                 mdi
            )]
        PushButton: add_btn:
            text = 'Add New MDI Window'
            clicked ::
                dyn_win.objects.append(MdiContent(num=count))
                main.count += 1
        MdiArea: mdi:
            Include: dyn_win:
                pass
```

MatthieuDartiailh:
Also add a MdiArea example and support for tiling and cascading of MdiWindows, but no way to manipulate the order of tiling/cascading. The documentation is also updated.

Closes #147

@frmdstryr @sccolbert feel free to give a look and comment.


frmdstryr:
Tested and works fine on 2.7 and 3.5.

I think it might be slightly cleaner to add an `arrangement` or `window_layout` attribute to the `MdiArea` set to an `Enum('default', 'tile', 'cascade')` instead of separate methods for each.

MatthieuDartiailh:
The point is that the user can always mess up the mdi layout and by calling the method he asks for a definite re-layout of the windows. If this was an attribute it should go `user-defined` as soon as the user manipulate a window which means capturing mouse event, which I would prefer to avoid.

frmdstryr:
Agree, that makes more sense.  

Could do a single `rearrange` or `reset` method that takes the layout as an argument, but the way you have now is consistent with Qt so it'd say it's good as is then.

MatthieuDartiailh:
wx uses the same terminology so I think we are safe here. I will try to ping Chris about that.

sccolbert:
This all looks good to me. I would maybe consider naming the methods `tile()` and `cascade()` though.

MatthieuDartiailh:
Sounds good, I will rename them ! Thanks for the feedback !

MatthieuDartiailh:
Even though I kind of liked the explicit names.

sccolbert:
I'll leave that up to you to pick what you think fits/feels best with the rest of the framework.

MatthieuDartiailh:
Looking around at the codebase (DockArea.save_layout) , I believe I do prefer the explicit names so I will merge as is. Thanks for the feedback.

Issue #266, opened 2018-05-30
Why is the import_hooks parse call not passing the source file name?


frmdstryr:
When the importer calls parse in [import_hooks.py](https://github.com/nucleic/enaml/blob/master/enaml/core/import_hooks.py#L373) there's an optional second argument [`filename`](https://github.com/nucleic/enaml/blob/master/enaml/core/parser/__init__.py#L38) that's not being used.

For syntax errors I get tracebacks missing the filename like:
```python-traceback
  File "Enaml", line 168
    Conditional
    ^
SyntaxError: invalid syntax
```

If I pass in the `file_info.src_path` then it shows it

```python-traceback

  File "/the/path/to/the/file.enaml", line 168
    Conditional
    ^
SyntaxError: invalid syntax

```

Is there a reason this isn't being used? The first commit to this repo is not using it https://github.com/nucleic/enaml/commit/892883d0252cbb5dcd4c529a69ebc6f8afa04001#diff-dfd53081b9a562110bd9f72ef0c04e15R350

MatthieuDartiailh:
Honestly I have no idea. @sccolbert do you remember the reason behind that choice ? If we can improve the debugging experience by passing the filename we should !

MatthieuDartiailh:
This can be closed, right ?

PR #279, opened 2018-07-01
Add more parser tests

frmdstryr:
Includes test for #266 

Also added test for a strange syntax error when using async def.


frmdstryr:
Going to break into separate PR's

frmdstryr:
@MatthieuDartiailh do you prefer the async tests go in a separate test module or should they be added to `test_function_definitions.py` and changed to use `validate_ast`?

MatthieuDartiailh:
Before taking a decision, could you explain a bit more the context of those tests ? Since you are more focusing on the function body and in particular async/await stuff a new module dedicated to those may make sense.

frmdstryr:
I'm building an analytics dashboard for a client using pandas, plotly, and enaml-web.  It loads data from a db using the await syntax. 

Something like

```python

async def fetch(view, request):
     # Build the query from request ...
     data = await db.fetch(query)
     # Load with pandas ...
     view.dataset = df

enamldef Page(BasePage):
    # Request handler checks if prepare exists and does  `await page.prepare(page, request)` 
    # then renders it and returns the result to the client
    attr prepare = fetch
    # Data
    attr dataset: pd.DataFrame
    # Rest of page does charts and stuff

```

In order to keep things organized, I want to put each view and related code in it's own separate enaml file.
It worked fine this way until I needed an `if` statement to build the query (hence the strange tests) so I currently have to push the fetching code into a separate py file and import it.  

It would be nice if the `func` keyword could be async so it's all encapsulated but this way works fine.

MatthieuDartiailh:
Thanks for the background. 
To me it looks a bit like you are mixing model and view here, but I am not familiar with enaml-web. Since you are really testing that async/await code works as expected in enaml, a new file is probably better, but I would use validate_ast since if something is wrong it is the AST we build (after we just rely on the Python compiler).
Honestly I have no experience with async so I have no idea what it would cost to create a Declarative coroutine, but if you feel up to it feel free to try to get something together.

PR #280, opened 2018-07-01
Test that the traceback has correct file path for syntax errors

frmdstryr:
For #266. Uses tmpdir.strpath to work on 2.7


tests/test_parser.py
```diff
@@ -0,0 +1,53 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013-2018, Nucleic Development Team.
```
MatthieuDartiailh: This copyright is wrong obviously.

PR #281, opened 2018-07-03
Add async tests

frmdstryr:


frmdstryr:
I think this is close to ready. 

Let me know what to do about the lexer3 hack I added. I'm not sure if some of that can be removed.

Also I can't get it to cover these https://codecov.io/gh/nucleic/enaml/src/fc1e73604b36a0fdfbbcccb406e697de8571576e/enaml/core/parser/parser36.py#L43...63

The 3.7 tests pass locally but Travis is still installing 3.6 because several dependencies haven't been built for 3.7 yet (pyqt, ply, matplotlib).

enaml/core/parser/lexer3.py
```diff
@@ -81,7 +81,7 @@ def make_token_stream(self):
 
         """
         token_stream = super(Python35EnamlLexer, self).make_token_stream()
-        token_stream = self.analyse_async(token_stream)
+        #token_stream = self.analyse_async(token_stream)
```
MatthieuDartiailh: The idea with this pass was to preserve the behavior of python that allows async and await as name in non async functions. I agree this is not critical. I will try to see if we can fix it. I would prefer to stay as close as possible to the promise that any valid Python is valid enaml.

enaml/core/parser/parser35.py
```diff
@@ -143,7 +143,7 @@ def p_async_with_stmt(self, p):
 
     def p_atom_expr3(self, p):
         ''' atom_expr : AWAIT atom '''
-        p[0] = ast.Await(value=p[1])
+        p[0] = ast.Await(value=p[2])
```
MatthieuDartiailh: Good catch

enaml/core/parser/parser36.py
```diff
@@ -32,28 +32,28 @@ class Python36EnamlParser(Python35EnamlParser):
     def p_comp_for1(self, p):
         ''' comp_for : FOR exprlist IN or_test '''
         super(Python36EnamlParser, self).p_comp_for1(p)
-        p[0][0].is_async = False
+        p[0][0].is_async = 0
```
MatthieuDartiailh: Sorry for that the name of the field got me confused. I will propose a patch at https://greentreesnakes.readthedocs.io/en/latest/nodes.html#comprehensions to make it clearer that this uses an int.

tests/core/parser/test_async.py
```diff
@@ -0,0 +1,128 @@
+
```
MatthieuDartiailh: Useless blank line

tests/core/parser/test_async.py
```diff
@@ -0,0 +1,128 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+import sys
+import ast
+from textwrap import dedent
+import pytest
+
+from enaml.core.parser import parse
+
+from .test_parser import validate_ast
```
MatthieuDartiailh: This looks quite neat.

.travis.yml
```diff
@@ -11,6 +11,7 @@ matrix:
     - env: PYTHON=3.4 QT_VERSION=4 IPY_VERSION=3
     - env: PYTHON=3.5 QT_VERSION=5 IPY_VERSION=5 QSCINTILLA=0
     - env: PYTHON=3.6 QT_VERSION=5 IPY_VERSION=6 QSCINTILLA=1
+    - env: PYTHON=3.7 QT_VERSION=5 IPY_VERSION=6 QSCINTILLA=1
```
MatthieuDartiailh: I am not sure we can claim support just yet. In particular I suspect that fast call related changes may need some special care.

MatthieuDartiailh: But we can keep it in.

frmdstryr: I was going to create a ticket to track support for 3.7, but I tested my applications and didn't have any issues.   

I dissembled some code and  it shows the new `LOAD_METHOD` and `CALL_METHOD` are used. 

Ex

```python

Python 3.7.0 (default, Jun 28 2018, 13:15:42) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dis

In [2]: from atom.api import *
   ...: class Test(Atom):
   ...:     x = Unicode()
   ...:     def bar(self):
   ...:         return self.foo()
   ...:     def foo(self):
   ...:         return self.x
   ...:     

In [3]: t = Test()

In [4]: dis.dis(t.bar)
  5           0 LOAD_FAST                0 (self)
              2 LOAD_METHOD              0 (foo)
              4 CALL_METHOD              0
              6 RETURN_VALUE

In [5]: dis.dis(t.foo)
  7           0 LOAD_FAST                0 (self)
              2 LOAD_ATTR                0 (x)
              4 RETURN_VALUE

In [6]: def baz():
   ...:     t.foo()
   ...:     

In [7]: dis.dis(baz)
  2           0 LOAD_GLOBAL              0 (t)
              2 LOAD_METHOD              1 (foo)
              4 CALL_METHOD              0
              6 POP_TOP
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE

```

Will the tracer need updated to handle these?

enaml/core/parser/lexer3.py
frmdstryr: Ok, yeah that makes sense. It should be left in for 3.5 and 3.6 but overridden to ignore it for 3.7 (in python 3.7 async and await are reserved keywords).

Some more tests should be added to check this then.

.travis.yml
MatthieuDartiailh: Yes I suspect so but I need to dig a bit more before being sure.

MatthieuDartiailh:
To cover the missing lines you need an async comprehension `[x async for x in y]`

enaml/core/parser/parser36.py
MatthieuDartiailh: And this is now merged.

enaml/core/parser/lexer3.py
MatthieuDartiailh: You need to reactivate this for the new tests to pass.

enaml/core/parser/lexer3.py
```diff
@@ -88,41 +88,25 @@ def analyse_async(self, token_stream):
...
-                self.handle_async_token(tok, async_depth)
-
-            elif async_depth and tok.type == 'DEDENT':
-                async_depth -= 1
+                next_token = next(token_stream)
+                if next_token.type != 'NAME':
+                    tok.type = 'NAME'
+                yield tok
+                yield next_token
+                continue
```
MatthieuDartiailh: This solution is indeed simpler than mine.

MatthieuDartiailh:
This looks good. I will merge. Thanks !

frmdstryr:
Thank you! 

I was also able to add support for `async func`  yesterday.  Just pushed to here https://github.com/nucleic/enaml/compare/master...frmdstryr:master

Is this something that would be good to include as well? I still need to add a test for it but it's working in my code.

MatthieuDartiailh:
Yes. Open a PR. Can you also support the => syntax ?

PR #285, opened 2018-07-04
Add support for declarative async func

frmdstryr:


MatthieuDartiailh:
I am not sure about it, but don't you need to modify the way the DeclarativeFunction is called to make it asynchronous ?

frmdstryr:
It's called the same, but the result needs to be passed to event loop to actually "run" the function.

An `async def` is just a generator function that yields a "future" or another generator that yields a future ( but they changed it to [not act like a generator](https://www.python.org/dev/peps/pep-0492/#differences-from-generators) for whatever reason).  See http://www.tornadoweb.org/en/stable/guide/coroutines.html for a good explanation.

I mainly want this for enaml-web, but using something like [qt5reactor](https://github.com/sunu/qt5reactor) would let you use this with [twisted](https://twistedmatrix.com/trac/) in a Qt (enaml) app to do all your networking io or serial port io all in one thread without blocking the UI.

MatthieuDartiailh:
Thanks for the interesting reading. Actually I got confused with what happens at the bytecode level. One case I am not sure how you are handling and that does not seem to be tested is the case of overriding an async function without using the async keyword or the other way around overriding with async something that lacks it. I am not sure of the behavior we want, but changing from non async to async looks like a bad idea to me.

frmdstryr:
>  changing from non async to async looks like a bad idea to me.

I agree, it can be bad.  But in some cases it's also useful. Also, python doesn't prevent this. 

Ex this works fine:

```python
In [44]: async def fetch():
    ...:     return True
    ...: class A:
    ...:     async def foo(self):
    ...:         result = await fetch()
    ...:         return result
    ...: class B(A):
    ...:     def foo(self):
    ...:         return fetch()
    ...: class C(B):
    ...:     async def foo(self):
    ...:         result = await fetch()
    ...:         return not result
    ...:     

In [45]: A().foo()
Out[45]: <coroutine object A.foo at 0x7f3636b28938>

In [46]: B().foo()
Out[46]:<coroutine object fetch at 0x7f3636acbeb8>

In [47]: C().foo()
Out[47]: <coroutine object C.foo at 0x7f3636b28eb8>

```
Converting a non async to an async will just make the function return a coroutine. Converting an async to non async doesn't do anything special, but the event loop will give an error if whatever returned isn't an "awaitable" object when it tries to "run" it.

I know some eventloops (at least twisted and tornado) have apis that will handle returning anything by wrapping any non async awaitable result into a a coroutine so you can use code from either case without caring which is used. 

I don't think we need to do anything special here.

frmdstryr:
Looking at the coverage, it might not need the changes to `block_compiler.py` since the parser just converts the `FuncDef` to an `AsyncFuncDef`.

MatthieuDartiailh:
I don't get why the visitor does not call the visit methods, it seems weird.
I actually checked that yes Python allows you to do anything you want, in term of overwriting methods. I also found this thread https://github.com/cython/cython/issues/2092 and I wonder if we should try to mark declarative async func as being coroutine based on this test.

frmdstryr:
> I don't get why the visitor does not call the visit methods, it seems weird.

Just realized this is because the test is only parsing, not compiling.  I removed it to double check and it does indeed require the changes to `block_compiler.py`.

> I wonder if we should try to mark declarative async func as being coroutine based on this test.

Nice find. Probably.

Do you suggest adding an `_is_coroutine` attribute or set the flag from the issue referenced there?

```python
def is_coroutine(function):
    """Returns True if the passed in function is a coroutine"""
    return function.__code__.co_flags & 0x0080 or getattr(function, '_is_coroutine', False)
```
 https://github.com/timothycrosley/hug/blob/develop/hug/introspect.py#L33-L35

If it's possible to set that flag without any implications, that'd be my vote.

frmdstryr:
Edit: I'm not sure that is correct. See https://github.com/python/cpython/blob/3327a2ddf1eaf8cc678f0219ba9edf7d632d1b20/Lib/types.py#L248-L266

When the async decl func is invoked it returns a coroutine object (from my working code).


```python
enamldef Page(Html): root:
    async func get(request):
        query = {}
        # etc...
````

```python
# In an async request handler
print(asyncio.iscoroutine(view.get))
print(asyncio.iscoroutinefunction(view.get))
f = view.get(request, *args, **kwargs)
print(asyncio.iscoroutine(f))
await f
```

outputs

```python
False
False
True
```

So it is only iscoroutinefunction that doesn't return the expected result.

MatthieuDartiailh:
Ok. I am not really familiar with asyncio so I will trust you on this one (but it seems to make sense).
Can you update the tests to include the compilation (I am not sure where to put this test however) ?

frmdstryr:
I guess that works.

MatthieuDartiailh:
Looks good. Merging !

PR #288, opened 2018-07-06
Add support for f-strings

MatthieuDartiailh:
@frmdstryr if you can review it would be great !


frmdstryr:
Would it be possible to also use cpythons fstring test? 

https://github.com/python/cpython/blob/3.6/Lib/test/test_fstring.py

enaml/core/parser/parser36.py
```diff
@@ -91,3 +88,30 @@ def p_annassign1(self, p):
...
+        p[0] = [repr(p[1])]
+
+    def p_atom_string_list2(self, p):
+        ''' atom_string_list : atom_string_list STRING '''
+        p[0] = p[1] + [repr(p[2])]
+
+    def p_atom_string_list3(self, p):
+        ''' atom_string_list : FSTRING '''
+        p[0] = ['f' + repr(p[1])]
+        print('parser')
```
frmdstryr: print left in

MatthieuDartiailh: Thanks. Will fix.

MatthieuDartiailh:
Using CPython tests may be interesting. But since we use ast.parse to generate the AST the main cases to check is the handling of escaped char otherwise we are pretty safe I would say.

MatthieuDartiailh:
Thanks for the reminder. I will do that when drafting the next release, which I should do soon. However I started working on atom and I would prefer to finish that first.

PR #291, opened 2018-07-17
Delete murpy.py theme

frmdstryr:
This is a duplicate of murphy.py.


MatthieuDartiailh:
This seems legitimate indeed.

PR #292, opened 2018-07-17
Add WebView example and another dock layout test

frmdstryr:
Want to add an example for the WebView to note that the QtWebEngineWidgets must be imported before the QtApplication is started.  qtpy currently clobbers the import error that mentions this and it takes time to find this.  It's imported in the example so it works when using enaml-run.

Also added a test for dock interactions. I would like to also get it to cover the some of the drag and drop code but QTest doesn't seem to support this very well.

### DIFF
--- a/examples/widgets/web_view.enaml
+++ b/examples/widgets/web_view.enaml
@@ -1,0 +1,27 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2018, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+""" 
+
+This example shows how to use a WebView. 
+
+<< autodoc-me >>
+"""
+from enaml.widgets.api import Window, Container, WebView
+
+
+# Note: QtWebEngineWidgets must be imported before the QtApplication is started
+# or it will throw an import error
+from enaml.qt import QtWebEngineWidgets
+
+
+enamldef Main(Window):
+    initial_size = (480, 960)
+    Container:
+        padding = 0
+        WebView:
+            url = 'https://enaml.readthedocs.org'
--- a/examples/widgets/web_view.enaml
+++ b/examples/widgets/web_view.enaml
@@ -9,13 +9,14 @@
 
 This example shows how to use a WebView. 
 
+Note: QtWebEngineWidgets must be imported before the QtApplication is started 
+on Qt 5.10+ or an ImportError will be thrown.
+
 << autodoc-me >>
 """
 from enaml.widgets.api import Window, Container, WebView
 
-
-# Note: QtWebEngineWidgets must be imported before the QtApplication is started
-# or it will throw an import error
+# Only needed for Qt 5.10+
 from enaml.qt import QtWebEngineWidgets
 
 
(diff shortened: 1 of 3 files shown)


examples/widgets/web_view.enaml
```diff
@@ -0,0 +1,27 @@
...
+""" 
+
+This example shows how to use a WebView. 
+
+<< autodoc-me >>
+"""
+from enaml.widgets.api import Window, Container, WebView
+
+
+# Note: QtWebEngineWidgets must be imported before the QtApplication is started
```
MatthieuDartiailh: I would move/duplicate this to/in the docstring, to make it more visible in the documentation. Furthermore, is this also true under qt4 or is this qt5 only ?

tests/test_dock_layout.py
```diff
@@ -69,3 +70,131 @@ def test_validation_dock_layout2(enaml_qtbot, enaml_sleep):
...
     assert globals() == glob
     enaml_qtbot.wait(enaml_sleep)
+    
+    
+def test_dock_area_interactions(enaml_qtbot, enaml_sleep):
+    """Test interations with the dock area.
+
+    """
+    # Since timers are used the sleep must be greater than the default
+    enaml_sleep = 300
```
MatthieuDartiailh: I am not sure I get why you change the sleep. Furthermore I would prefer to use a conditional (either `or` or an inline `if ... else`) to avoid overwriting the user decision (if I want to look at the interaction I may set the time to 1000.

tests/test_dock_layout.py
```diff
@@ -69,3 +70,131 @@ def test_validation_dock_layout2(enaml_qtbot, enaml_sleep):
...
+        enaml_qtbot.mouseClick(tb._link_button, Qt.LeftButton)
+        enaml_qtbot.wait(enaml_sleep)
+        
+        # Restore it
+        op = InsertTab(item=di.name, target='item_1')
+        dock.update_layout(op)
+        enaml_qtbot.wait(enaml_sleep)
+        
+        # Drag it around
+        # TODO: this doesn't currently drag the window. It seems like QTest
```
MatthieuDartiailh: Since this does not work, is it worth keeping it around ?

frmdstryr: It can be removed, but I was hoping someone had more insight on how to do this.  If you add a long sleep and drag it around manually it catches about another 500 lines of coverage.

tests/test_dock_layout.py
frmdstryr: Would a `enaml_sleep = max(300, enaml_sleep)` be sufficient?

examples/widgets/web_view.enaml
frmdstryr: It appears to be with PyQt 5.10+. With 5.9.2 and the import is not needed.  With 5.10.1 and 5.11.2 and it is needed.

```python
  File "<root>/miniconda3/envs/occ/lib/python3.6/site-packages/qtpy/QtWebEngineWidgets.py", line 22, in <module>
    from PyQt5.QtWebEngineWidgets import QWebEnginePage
ImportError: QtWebEngineWidgets must be imported before a QCoreApplication instance is created
```

Should the version be checked and imported only if needed?

```
from enaml.qt.QtCore import __version_info__ as qt_version_info
if qt_version_info >= (5, 10):
    from enaml.qt import QtWebEngineWidgets
```

tests/test_dock_layout.py
MatthieuDartiailh: Yes indeed.

tests/test_dock_layout.py
MatthieuDartiailh: Ok then it is worth investigating. But some quick searches on google did not point to any solution.

MatthieuDartiailh:
This is nice to have and improve significantly the coverage. Thanks

PR #294, opened 2018-07-23
Documentation improvements.

Julian-O:
The main change here is the addition of a new example file, which demonstrates how the various date-related widgets look and work, and documentation to match.

While doing that work, I made a number of other more minor changes, including reducing the Sphinx warnings, which included exposing another example that had been hidden.


MatthieuDartiailh:
You should update the example doc generator script rather than each example separately.

MatthieuDartiailh:
The new format look nice, as mentioned in my previous message you should update the template in docs/source/examples/example_doc_generator.py

Julian-O:
Oh, I am so glad I submitted the example template early. That will save me so much time!

MatthieuDartiailh:
Actually the enaml pygments plugin exists. It is not on PyPI but you can find it here https://github.com/nucleic/enaml/tree/master/tools/pygments and there is a conda package for it on the exopy channel that we use for building the docs. So please revert that change.

Julian-O:
@MatthieuDartiailh: Any clues how to (directly) install that plugin?

I run: `python tools\pygments\setup.py install`. During the installation I get the following error:

    Processing enamllexer-0.0.0-py3.6.egg
    creating c:\stores\caches\virtualenv\enaml7\lib\site-packages\enamllexer-0.0.0-py3.6.egg
    Extracting enamllexer-0.0.0-py3.6.egg to c:\stores\caches\virtualenv\enaml7\lib\site-packages
      File "c:\stores\caches\virtualenv\enaml7\lib\site-packages\enamllexer-0.0.0- py3.6.egg\enaml\core\byteplay\byteplay2.py", line 183
    raise ValueError, "Opcode stack behaviour depends on arg"
                        ^
    SyntaxError: invalid syntax

(There is a Python 2 and Python 3 version of this code - I don't know why the Python 2 version is being parsed here.)

Despite that, it appears to complete installation.

Then, when I run `make html` in the docs directory, it gives:

    Exception occurred:
      File "c:\[...]\site-packages\pkg_resources\__init__.py", line 2330, in resolve
        module = __import__(self.module_name, fromlist=['__name__'], level=0)
    ModuleNotFoundError: No module named 'enamllexer'
    The full traceback has been saved in C:\[...]\Temp\sphinx-err-lx0pxrhb.log, if you want to report the issue to the developers.
    Please also report this if it was a user error, so that a better error message can be provided next time.
    A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!

It looks like it didn't install correctly. Am I doing it right?

MatthieuDartiailh:
I cannot check at the moment, but I think it would be better to cd into the folder `tools/pygments`. In a number of case running setup.py works best when setup.py is run from the current directory (not sure of my wording here).

Julian-O:
I inspected the failed Travis test. Intermittent network error during Travis CI start up - it's not a reflection on the code being submitted.

Julian-O:
Just a quick status summary:

* This is PR ready for review - and merging, if the review is successful.
* I haven't been able to test the recent (minor) change to revert the language to Enaml, as I am still getting no module found. (Installing from the current directory seems to have resolved the Python 2 code being interpreted exception though.)
* The Travis error can be ignored.

docs/source/examples/example_doc_generator.py
```diff
@@ -14,18 +14,22 @@
...
 import shutil
 
 from atom.api import Atom, Unicode, Value
 import enaml
 from enaml.qt.qt_application import QtApplication
 from enaml.qt.QtWidgets import QApplication
 from enaml.qt.QtTest import QTest
-from enaml.application import timed_call
 
+import sys # TODo: Remove
```
frmdstryr: Should have an extra newline here for  and between the save_snapshot_of_module, extract_docstring and clean_docstring functions below to make PEP-8 linters happy.

frmdstryr:
Looks nice! 

Should the doc string be removed from the example code? Not a big deal but currently it's repeated.

docs/source/examples/example_doc_generator.py
```diff
@@ -82,69 +115,68 @@ def generate_example_doc(app, docs_path, script_path):
...
+    # Restore Python path.
+    sys.path = old_python_path
+
+    with open(os.path.join(script_path)) as fid:
+        script_text = fid.read()
+
+    docstring = clean_docstring(extract_docstring(script_text))
+
+    screenshot_rst = '' if not snapshot_success else ("""
+Screenshot
```
frmdstryr: Using `textwrap.dedent` or separate global variables for the template strings might make this easier to read.

MatthieuDartiailh:
Thanks for giving this a look @frmdstryr. I may take some time to review since I just moved (from France to NY) and because I want to take the time to rebuild the docs in a clean env and document how to do so.

frmdstryr:
Take all the time you need and welcome to the US :+1: .

Julian-O:
Moving countries is hard.  Good luck with it!

Here are my notes on how I generated the docs, that you might find useful for documenting the process:

*  Create new virtualenv.
* Get the updating tools up to date
    * `python -m pip install -I pip`
    * `pip install -I setuptools`
* Pre-install requirements that the Enaml setup.py won't.
    * `pip install https://github.com/nucleic/atom/tarball/master pyqt5!=5.11.2`
* Install requirements that Enaml doesn't need but the examples do.
    * `pip install numpy qtconsole matplotlib qscintilla vtk`
* Ensure VisualStudio/c compiler is installed/on the path.
* Install Enaml
    * `python setup.py develop`
    * `cd tools\pygments\`
    * `python setup.py install`
    * `cd ..\..`
* Build the examples and documents.
    * `cd docs`
    * `make examples html`

These instructions still lead to errors with the Pygments lexer, an error with importing a vtk library, and one other import error (QSci?) that I never chased down.

docs/source/examples/example_doc_generator.py
Julian-O: Thank you for the introduction to textwrap.dedent(). I never knew that feature was hidden in that module! Hope to use it every time I am embedding long strings now.

frmdstryr: @MatthieuDartiailh introduced it to me so thank him :+1:

MatthieuDartiailh:
Playing around a bit to try to get the lexer to run it looks like it can fail on some construct. See for example in layout.rst the block line 123, the argument of align fail to be processed, I checked that it is not related to the presence of comments but apparently only to the presence of the parentheses.

MatthieuDartiailh:
I took the liberty to push some fixes. The enaml lexer for pygments is now correct as far as I can tell. The docs build without any warning anymore and I added instruction about building the docs and rebuilt the examples.

docs/README.rst
```diff
@@ -0,0 +1,16 @@
...
+Building the docs involve the following steps:
+
+- install enaml and its required dependencies: ply, qtpy, atom, kwisolver
+- install sphinx
+- cd into the docs folder and run `make html`
+
+Rebuilding the example pages and screenshot involve the following steps:
+
+- install enaml and its required dependencies: ply, qtpy, atom, kiwisolver
+- install in addition: pyqt, numpy, matplotlib, itpython, qtconsole, vtk,
```
frmdstryr: Typo with `itpython`

MatthieuDartiailh:
I agree about vtk @Julian-O. I will update the code but I won't have time to test. The echo is in line with the other command so I will leave it.
Thanks @frmdstryr. I will fix it in my next commit. I will also add instructions about installing the pygments lexer.

MatthieuDartiailh:
@Julian-O @frmdstryr I will merge tomorrow or Monday in case you have any further comments.

Julian-O:
I'm happy for the merge. Thanks.

tools/pygments/setup.py
```diff
@@ -3,6 +3,7 @@
 
 setup (
   name='enamllexer',
+  version='0.1.0'
```
frmdstryr: Missing comma

MatthieuDartiailh: indeed ...

frmdstryr:
I clicked next through everything. Overall this is looking great. A few final notes:

1. Highlighting in the tutorial's doesn't seem to be working
2. Using osx for the menu bar example probably doesn't make sense (as the it goes to the system menu and the screenshot is a blank window)
3. Would it make more sense to include the path in the command line relative to the root when cloning? I think users will typically use `enaml-run examples/<type>/<example>.enaml`  instead of changing to each directory

I'd say it's fine to pull as is though.

Julian-O:
@frmdstryr Re: 3

While the examples folder path location is pretty clear for developers, end-users will install enaml via pip, so the location of the examples folder will be hidden (somewhere in site-packages, perhaps zipped up).

Instead, the instructions instruct you to download a new copy of the enaml file to your favourite temporary location, and run it directly from there.

MatthieuDartiailh:
I "fixed" the highlighting in the tutorials (part of it, in some because there are incomplete snippets the first widget is not highlighted).
In the future it may be nice to have screenshots under the different OSes (demonstrate a native look) but for the time being the OSX screenshots are fine (I mean only one of them is kind of problematic no ?).
I don't have a strong opinion on 3 but I believe that people going so far as cloning the repo can figure out they can call the examples without going into the directory first.

Thanks to both of you for your work and your feedback. Merging.

PR #298, opened 2018-08-02
Updates lexer/parser to allow using def for declarative functions

frmdstryr:
Retains backwards compatibility by accepting either `func` or `def` when defining a declarative func. It does not do anything with overrides.

I tested it by using find and replace in an existing project with both sync and async functions.

### DIFF
--- a/enaml/core/parser/base_lexer.py
+++ b/enaml/core/parser/base_lexer.py
@@ -581,7 +581,30 @@ class BaseEnamlLexer(object):
         token_stream = self.annotate_indentation_state(token_stream)
         token_stream = self.synthesize_indentation_tokens(token_stream)
         token_stream = self.add_endmarker(token_stream)
+        token_stream = self.annotate_enamldef_state(token_stream)
         return token_stream
+
+    def annotate_enamldef_state(self, token_stream):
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
@sccolbert could you comment on this ? Personally I believe it makes sense but since it was not your choice I would like your opinion. 
@frmdstryr could you please use a feature branch based on nucleic/master to avoid the merge commits ? also this will require a documentation update.

frmdstryr:
Thanks, I updated to a feature branch.  Will wait to update the docs until it's determined whether or not we want to do this.  Perhaps a discussion on the google group is better? 

I was thinking about this more and was also wondering why an override was not done using another keyword? Ex

```

enamldef MyWidget(Container):
    func do_stuff():
       # ...
enamldef SomeWidget(MyWidget):
    override do_stuff():
        # ...
```

Not suggesting the change, just interested in why the alternate syntax was chosen, I assume it was just easier to do however it was implemented at the time.

MatthieuDartiailh:
@sccolbert could you comment on the principle of this update ?

MatthieuDartiailh:
ping @sccolbert

MatthieuDartiailh:
Fair enough.

Issue #287, opened 2018-07-05
Python 3.7 support 


MatthieuDartiailh:
Full support for Python 3.7 requires to at least check the following items:
- [x] make sure we treat async/await are treated as full keyword (PR #281)
- [x] check the impact of LOAD_METHOD and CALL_METHOD new opcodes
- [ ] make sure both Travis and Appveyor do run the tests on Python 3.7 (conda sometimes downgrade to 3.6)

MatthieuDartiailh:
This is blocked by atom support for 3.7

odinlake:
enaml/enaml/core/byteplay/byteplay2.py contain raise statements that are not valid py3. I noticed this in the output of "setup.py install" while it was extracting a .egg.

MatthieuDartiailh:
This can be safely ignored since this file is not used under Python 3. Actually the last release of enaml support and is tested under Python 3.7, I just forgot to close that issue and mention it in the release notes !

PR #300, opened 2018-08-06
Run Appveyor tests on Python 3.7.

Julian-O:
Addresses one of the requirements for #287.

* Miniconda doesn't have a Python 3.7 version yet.
* PyQt5 5.11.2 was improperly built for Python 3.5. Looks like they fixed the problem without incrementing the version number.
* Minor spelling and spacing improvements.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,31 +13,39 @@ environment:
         #    Enaml supports Python 2.7 and also 3.4 onwards.
         #    PyQt4 only available for Python <3.6.
         #    PyQt5 only available for Python >=3.5.
-        #       PyQt5 version 5.11.2 doesn't run on Python 3.5.
         #    PySide1 only available for 2.6 <= Python <=3.4. Wheels only for 32-bit.
-        
+        #    Miniconda does not run on Python 3.7 yet.
+
         # Strategy:
         #    Test each 64-bit Python version, but only the latest known patch.
-        #    Test the oldest and the latest 32-bit Python versions release. 
-        #    Test with PyQt4 where available, otherwise PyQT5 in each version.
+        #    Test the oldest and the latest 32-bit Python versions release.
+        #    Test with PyQt4 where available, otherwise PyQt5 in each version.
         #    Test QScintilla on latest Python version available.
-        #    Test Pyside1 in latest Python version it runs on.
+        #    Test PySide1 in latest Python version it runs on.
         #        - Failure is allowed and won't block PRs.
-        #    Test Pyside2 in latest Python version available
+        #    Test PySide2 in latest Python version available
         #        - Failure is allowed and won't block PRs.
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
I must say I am considering re-writing the appveyor config to sinmly use the latest miniconda and then create a dedicated environment to run the tests. It makes easier to support 3.7 and we avoid having to deal with old distributions. WDYT ?

Julian-O:
It depends what your goals are.

The Travis file mainly relies on Miniconda, which is a great way to quickly get a stable environment and actually unit-test the Python.

But one of the motivating factors of 
  * re-writing the installation documentation,
  * producing the Appveyor testing, and 
  * relying as much as possible on PIP in the appveyor testing was the initial difficulty I had installing Enaml on Windows.

So, as much as anything, I wanted the Appveyor script to test that Enaml and its dependencies could be *installed*. 

I think you should test the platforms you claim to (fully) support. Saying you no longer fully support Python 3.5 and 3.6 is an option for the maintainers. It is a question of balancing effort versus target audience.

----

Some related questions:

* Is it time to stop fully supporting PyQt4? Then you could simply install all the dependencies straight from running setup.py.
* Should CI testing include a "generate examples and documentation" build?
* If so, should there be a target in setup.py for documentation dependencies?

MatthieuDartiailh:
We still support 2.7 (at least for another year) so we cannot completely drop pyqt4. Dropping 3.4 will become reasonable once we consolidate Python 3.7 support (at the time being atom is still broken on 3.7).
Typically those platform are harder to test without miniconda but I am fine with testing (3.5 if possible) 3.6 and 3.7 using only pip (I believe this is actually a good idea).
I usually keep an eye on ReadTheDocs for the documentation, but adding one build for docs may make sense. I do not think we need to build the examples since those are tested but now that vtk is on conda and works we should add it.
Given that appveyor is slow, I suggest we focus on the key tests (not all examples) for Appveyor, and run using conda on 2.7 and 3.4 since it will be easier and running fully using pip on the 3.5, 3.6 and 3.7. I will try to take care of the doc build and vtk on Travis.

Julian-O:
Re: PyQt4: That makes complete sense.

Re: Miniconda versus PIP: That makes sense too. I'd merge this change now, and make it a separate (perhaps low priority?) plan.

Re: Docs: I am thinking mainly about feedback/review so you can see that changes (whether to .rst files, example .enaml files, or others) have been properly reflected in the generated comments. Less concerned about automated generation for ReadTheDocs, although I guess it is a possibility. Again, I see this as icing on the cake, not a priority.

Re: Travis versus AppVeyor: I hadn't noticed the speed difference, although I bet I will now it has been pointed out! Doing docs and vtk on Travis only makes sense.

In summary, I agree with everything you've said. I recommend approving this change as a step towards Py3.7 support, and making the other items low-priority future plans.

PR #303, opened 2018-08-20
Add CloseEvent to DockItem

bburan:
This will allow DockItem widgets to perform an action in response to the titlebar close button being clicked (and ignore this event if desired).

### DIFF
--- a/enaml/qt/qt_dialog.py
+++ b/enaml/qt/qt_dialog.py
@@ -7,8 +7,8 @@
 #------------------------------------------------------------------------------
 from atom.api import Typed, atomref
 
-from enaml.widgets.dialog import ProxyDialog
 from enaml.widgets.close_event import CloseEvent
+from enaml.widgets.dialog import ProxyDialog
 
 from .QtCore import Qt
 from .QtWidgets import QDialog
--- a/enaml/qt/qt_dialog.py
+++ b/enaml/qt/qt_dialog.py
@@ -8,7 +8,7 @@
 from atom.api import Typed, atomref
 
 from enaml.widgets.dialog import ProxyDialog
-from enaml.widgets.window import CloseEvent
+from enaml.widgets.close_event import CloseEvent
 
 from .QtCore import Qt
 from .QtWidgets import QDialog
--- a/enaml/qt/qt_window.py
+++ b/enaml/qt/qt_window.py
@@ -10,7 +10,8 @@ import sys
 from atom.api import Typed, atomref
 
 from enaml.layout.geometry import Pos, Rect, Size
-from enaml.widgets.window import ProxyWindow, CloseEvent
+from enaml.widgets.window import ProxyWindow
+from enaml.widgets.close_event import CloseEvent
 
 from .QtCore import Qt, QPoint, QRect, QSize
 from .QtGui import QIcon
--- a/enaml/qt/qt_window.py
+++ b/enaml/qt/qt_window.py
@@ -10,8 +10,8 @@ import sys
 from atom.api import Typed, atomref
 
 from enaml.layout.geometry import Pos, Rect, Size
-from enaml.widgets.window import ProxyWindow
 from enaml.widgets.close_event import CloseEvent
+from enaml.widgets.window import ProxyWindow
 
 from .QtCore import Qt, QPoint, QRect, QSize
 from .QtGui import QIcon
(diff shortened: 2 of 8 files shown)


sccolbert:
Fine with the concept. Not a fan of the circular _proxy_ref thing. There
might be a cleaner way to express that, like using an event filter on the
qwidget.

I'm at jupytercon this week, so let's leave this open for a bit until I can
take a closer look.

MatthieuDartiailh:
Sure @sccolbert . The proxy_ref trick is similar to what you did on the QtWindow (but not at the same place and you used an atomref), hence why it was not really bothering to me (but I will have to check what atomref does...).

sccolbert:
atomref is a more efficient weakref for atom objects

bburan:
Are you suggesting:

```
class DockItemEventFilter(QObject):

    def __init__(self, declaration, *args, **kwargs):
        self.declaration = declaration
        QObject.__init__(self, *args, **kwargs)

    def eventFilter(self, object, event):
        if event.type() == QEvent.Close:
            close_event = CloseEvent()
            self.declaration.closing(close_event)
            if not close_event.is_accepted():
                event.ignore()
                return True
        return QObject.eventFilter(self, object, event)
```
Then doing the following in `QtDockItem`:

```
    def create_widget(self):
        """ Create the underlying QDockItem widget.

        """
        self.widget = QCustomDockItem(self.parent_widget())
        self.event_filter = DockItemEventFilter(self.declaration)
        self.widget.installEventFilter(self.event_filter)


```

sccolbert:
Yes.

However it might be worth looking at atomref if that's what I did for the
Window class.

MatthieuDartiailh:
ping @sccolbert 

The implementation now matches the one used for windows. To me this is good to go. Let me know If you think it is worth re-visiting the whole idea using event filters.

enaml/qt/qt_window.py
```diff
@@ -10,7 +10,8 @@
 from atom.api import Typed, atomref
 
 from enaml.layout.geometry import Pos, Rect, Size
-from enaml.widgets.window import ProxyWindow, CloseEvent
+from enaml.widgets.window import ProxyWindow
+from enaml.widgets.close_event import CloseEvent
```
sccolbert: `close_event` should be imported before `window` to maintain sort order. /pedantry

enaml/widgets/close_event.py
```diff
@@ -0,0 +1,35 @@
+from atom.api import Atom, Bool
```
sccolbert: This file needs a copyright header.

enaml/widgets/window.py
```diff
@@ -15,6 +15,7 @@
 from enaml.layout.geometry import Pos, Rect, Size
 
 from .container import Container
+from .close_event import CloseEvent
```
sccolbert: This should be imported before `.container` to maintain sort order.

enaml/widgets/close_event.py
MatthieuDartiailh: This is always a bit unclear to me but can the copyright mention 2013 when the file was created only in 2018 ?

sccolbert:
ideally, all files should be updated to current year.

sccolbert:
some of the imports in this PR are still in the wrong order

enaml/qt/qt_dialog.py
```diff
@@ -8,7 +8,7 @@
 from atom.api import Typed, atomref
 
 from enaml.widgets.dialog import ProxyDialog
-from enaml.widgets.window import CloseEvent
+from enaml.widgets.close_event import CloseEvent
```
sccolbert: Almost there. This is the last one that's out of order.

sccolbert:
:shipit:

MatthieuDartiailh:
Thanks @bburan this is a welcome addition. Thanks @sccolbert for the reviews.

bburan:
@MatthieuDartiailh My colleague was trying to test out Enaml by cutting-and-pasting examples from readthedocs. He couldn't get the dock_area.enaml to work because I added a closing event to the DockItem (to demo the addition in this PR) and the current conda channels do not yet reflect master (but readthedocs.org does). Do we want to worry about this detail? My coworker was getting a bit frustrated testing it out before he contacted me.

MatthieuDartiailh:
I probably should compile the docs for released versions (never had time to set that up). And make sure people notice they are looking at the development version docs.  Thanks for reporting.

MatthieuDartiailh:
Sadly for the last release the docs are broken ... I will add doc building to the CI and hopefully starting with the next release the stable docs will be available.

PR #306, opened 2018-08-30
examples: update observer function name so it gets picked up by Atom

nasefbasdf:
I tried it with the updated function name and it works. Not sure if I'd need to do anything else.


MatthieuDartiailh:
Thanks @nasefbasdf

PR #316, opened 2018-09-19
Add pickle support for enaml's Color

frmdstryr:


MatthieuDartiailh:
This looks good and based on https://docs.python.org/3/c-api/typeobj.html I believe you are right in changing tp_name (the same argument also applies on Python 2). Since it is so could you do the same for all other extension types ?
I will proof-read several times since I am never at ease when touching the C part of the code.

enaml/src/colorext.cpp
```diff
@@ -220,11 +241,6 @@ struct module_state {
 };
 
 
-static PyMethodDef
-colorext_methods[] = {
```
sccolbert: You can't delete this. This is the methoddef for the module, not the class. Look at line 248 in the original file. You need to rename this object to something else and update the moduledef.

sccolbert: Or rather, don't delete this, and rename your methoddef object for the class to something else.

enaml/src/colorext.cpp
```diff
@@ -160,9 +160,30 @@ Color_getset[] = {
 };
 
 
+static PyObject*
+Color__reduce__( Color* self, void* context )
+{
+    PyObject* cls = PyObject_Type(reinterpret_cast<PyObject*>( self ));
+    if ( !cls )
+        return 0;
+    uint32_t a = ( self->argb >> 24 ) & 255;
```
sccolbert: Can you use `0xFF` instead of `255`. I use hex notation when performing bit-masking, since it makes it a bit more clear what's going on.

frmdstryr: Agree. Updated.

enaml/src/colorext.cpp
MatthieuDartiailh: The point of @sccolbert was, I believe, that you are mixing up the methods of the class object (that you need to add) and the ones of the module object which should not change and hence why you must not delete that piece of code.

sccolbert: ^ this

MatthieuDartiailh:
Since @sccolbert approves I will merge and try to keep in mind that we should update the tp_name slots of the other objects.

enaml/src/colorext.cpp
```diff
@@ -160,9 +160,30 @@ Color_getset[] = {
 };
 
 
+static PyObject*
+Color__reduce__( Color* self, void* context )
+{
+    PyObject* cls = PyObject_Type(reinterpret_cast<PyObject*>( self ));
```
sccolbert: `PyObject_Type` returns a new reference to the type object. The problem is that `Py_BuildValue` also increments the reference of the object passed with the `O` flag. So, this method is leaking a reference to the type object.

The proper way to do this (which only borrows a reference) would be:
`PyObject* cls = reinterpret_cast<PyObject*>( self->ob_type );`

MatthieuDartiailh: Nice catch !

frmdstryr: I had tried something like this originally but it is giving an error
```
enaml/src/colorext.cpp: In function ‘PyObject* Color__reduce__(Color*, void*)’:
enaml/src/colorext.cpp:166:56: error: ‘struct Color’ has no member named ‘ob_type’
     PyObject* cls = reinterpret_cast<PyObject*>( self->ob_type );
                                                        ^
error: command 'gcc' failed with exit status 1
```

This works

```cpp

static PyObject*
Color__reduce__( Color* self, void* context )
{
    PyObject* obj = reinterpret_cast<PyObject*>( self );
    if ( !obj )
        return 0;
    uint32_t a = ( self->argb >> 24 ) & 0xFF;
    uint32_t r = ( self->argb >> 16 ) & 0xFF;
    uint32_t g = ( self->argb >> 8 ) & 0xFF;
    uint32_t b = self->argb & 0xFF;
    return Py_BuildValue( "O (iiii)", obj->ob_type, r, g, b, a);
}



```

enaml/src/colorext.cpp
```diff
@@ -163,14 +163,14 @@ Color_getset[] = {
 static PyObject*
 Color__reduce__( Color* self, void* context )
 {
-    PyObject* cls = PyObject_Type(reinterpret_cast<PyObject*>( self ));
-    if ( !cls )
+    PyObject* obj = reinterpret_cast<PyObject*>( self );
```
sccolbert: This cast will never fail. No need to check for `if( !obj )`.

frmdstryr: Thanks amended the commit to remove it.

sccolbert:
:shipit: Thanks!

MatthieuDartiailh:
ping @frmdstryr could you update the changelog ?

frmdstryr:
Updated

MatthieuDartiailh:
Thanks

PR #323, opened 2018-10-21
Fix order of text in generated .pdf file

rogererens:
Currently, in the generated .pdf file, the toctree gets expanded before the reader's guide of the Getting Started chapter (i.e. the reader's guide ends up on pp. 18 and 19).
I'm not well-versed into sphinx, so I don't know if this change actually does fix that order...

### DIFF
--- a/docs/source/dev_guides/index.rst
+++ b/docs/source/dev_guides/index.rst
@@ -10,13 +10,6 @@ variety of topics ranging from language features and syntax rules to
 application development and best practices. For the implementation details of
 the framework see the :ref:`arch_ref`.
 
-.. toctree::
-    :hidden:
-
-    stylesheets
-    languagebasedtools
-    workbenches
-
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
I am not sure I follow what you are trying to fix. Could you check that this fixed your issue by building locally (`make pdf` from the docs folder) ?

rogererens:
Sorry, I don't have the tool chain set up (e.g. no make available).
Maybe the screen shot shows what I was aiming to improve:
image

The non-numbered sections 'Introduction', 'Installation', 'Anatomy of Enaml' at the bottom of page 18 (and the sections on page 19) are currently at the end Chapter 1, after 1.5 'Constraints Layout'. I believe they should be shown before 1.1 'Introduction', on page 3.

MatthieuDartiailh:
I confirmed that this fixes your issue and I applied the same fix to the developer guide. I will wait for Travis and merge.

PR #328, opened 2018-10-25
Keep ctrl-c working with enaml-run

frmdstryr:


MatthieuDartiailh:
Could you comment a bit on this please ? and also update the changelog ?

MatthieuDartiailh:
Reading a bit about signals and Qt, this makes sense. But it also means that whenever using this you get an abrupt exit (finally clauses are not honored). Since enaml-run is mostly meant for small tests it should be fine. I will merge sometimes next week in case somebody wishes to comment.

MatthieuDartiailh:
Thanks for the test.

PR #331, opened 2018-11-08
Update import hook logic for Python 3

MatthieuDartiailh:
Python 3 introduced find_spec, create_module, exec_module. The old API generate DeprecationWarning. We should also look into a more 'complete' implementation (get_code, is_package, ...) since the lack of those may bug tools like cxFreeze or similar packagers.

This needs tests !

### DIFF
--- a/enaml/core/import_hooks.py
+++ b/enaml/core/import_hooks.py
@@ -186,8 +186,6 @@ class AbstractEnamlImporter(with_metaclass(ABCMeta, object)):
         If a module already exisist in sys.path, the existing module is
         reused, otherwise a new one is created.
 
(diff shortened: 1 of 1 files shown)


frmdstryr:
Looks good to me.

MatthieuDartiailh:
Thanks for the review @frmdstryr .

frmdstryr:
The  "This method is used only in Python 3.4+" comments somewhat misleading because it's being called from load_module

MatthieuDartiailh:
Indeed, it was true at the beginning but not anymore.

MatthieuDartiailh:
Fixed

PR #346, opened 2019-04-03
Add button group widget

djmattyg007:
This is a proxy for Qt's QButtonGroup class, which allows manual
management of button groupings. This alleviates the restriction of
requiring all buttons to be within the same container to ensure that
only one button in the desired grouping can be selected at a time.

I don't think this is ready to merge in its current state. The
requirement to manually hook up the button to the button group in the
'activated' event handler is very clunky, but I'm not sure at present
of a better way of handling this.

I'm also not sure that it makes sense to have the button group as an
actual widget, but I wasn't sure of a better way to handle this either.
The Timer widget (a proxy for the QTimer class) appears to work in the
same way that this patch currently does.

I'm pretty sure there is some other documentation that needs updating,
such as anything that claims the only way to implement exclusitivity is
through a common container parent. I'd appreciate some guidance on this,
and on if any additional testing is required.

I also wanted to say thank you for such an amazing piece of work. I
found Enaml just three weeks ago, and it's been an absolute joy to work
with. I didn't think it was possible for desktop application development
to be so easy and fun. I know a large part of the software development
industry is moving towards the web, but there are absolutely still plenty
of important use cases where a desktop application makes more sense, so
it's very nice to see toolkits like this still being developed.

### DIFF
--- a/enaml/qt/qt_abstract_button.py
+++ b/enaml/qt/qt_abstract_button.py
@@ -33,6 +33,7 @@ class QtAbstractButton(QtControl, ProxyAbstractButton):
     #: A reference to the widget created by the proxy
     widget = Typed(QAbstractButton)
 
+    #: A reference to the group this button is part of.
     _current_group = Typed(ButtonGroup)
 
     #: Cyclic notification guard. This a bitfield of multiple guards.
@@ -118,6 +119,14 @@ class QtAbstractButton(QtControl, ProxyAbstractButton):
             self.widget.setIconSize(QSize(*size))
 
     def set_group(self, group):
+        """Set the group this button belongs to.
+
+        Parameters
+        ----------
+        group : ButtonGroup
+            ButtonGroup declaration to which this button should be added.
+
+        """
         if self._current_group:
             self._current_group.proxy.remove_button(self.declaration)
         group.proxy.add_button(self.declaration)
--- a/enaml/qt/qt_abstract_button.py
+++ b/enaml/qt/qt_abstract_button.py
@@ -8,6 +8,7 @@
 from atom.api import Int, Typed
 
 from enaml.widgets.abstract_button import ProxyAbstractButton
+from enaml.widgets.button_group import ButtonGroup
 
 from .QtCore import QSize
 from .QtGui import QIcon
@@ -31,6 +32,8 @@ class QtAbstractButton(QtControl, ProxyAbstractButton):
     """
     #: A reference to the widget created by the proxy
     widget = Typed(QAbstractButton)
+
+    _current_group = Typed(ButtonGroup)
 
     #: Cyclic notification guard. This a bitfield of multiple guards.
     _guard = Int(0)
@@ -56,6 +59,8 @@ class QtAbstractButton(QtControl, ProxyAbstractButton):
             self.set_icon(d.icon)
         if -1 not in d.icon_size:
             self.set_icon_size(d.icon_size)
+        if d.group:
+            self.set_group(d.group)
         self.set_checkable(d.checkable)
         self.set_checked(d.checked)
         widget = self.widget
@@ -111,6 +116,12 @@ class QtAbstractButton(QtControl, ProxyAbstractButton):
         """
         with self.geometry_guard():
             self.widget.setIconSize(QSize(*size))
+
+    def set_group(self, group):
+        if self._current_group:
+            self._current_group.proxy.remove_button(self.declaration)
+        group.proxy.add_button(self.declaration)
+        self._current_group = group
 
     def set_checkable(self, checkable):
         """ Sets whether or not the widget is checkable.
--- a/enaml/qt/qt_button_group.py
+++ b/enaml/qt/qt_button_group.py
@@ -1,3 +1,10 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2019, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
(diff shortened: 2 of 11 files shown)


djmattyg007:
I'm not sure how to resolve the issue with the Travis CI build.

sccolbert:
Did you try my suggestion in the previous thread about using the `name` attribute to setup a button group?

sccolbert:
Assuming that the `name` approach does not work, I think what I would rather see is a `group` attribute added to the AbstractButton class, and then the toolkit implementations add themselves to that group whenever it is assigned. This is opposed to having to do it procedurally via the `activate` event as you've done here.

sccolbert:
i.e.

```
enamldef Main(Window):
    Container:
        ButtonGroup: btn_group1:
            pass
        ButtonGroup: btn_group2:
            pass
        GroupBox:
            title = "Group 1"
            RadioButton:
                text = "Button Group 1"
                group = btn_group1
            RadioButton:
                text = "Button Group 2"
                group = btn_group2
        GroupBox:
            title = "Group 2"
            RadioButton:
                text = "Button Group 1"
                group = btn_group1
            RadioButton:
                text = "Button Group 2"
                group = btn_group2
```

djmattyg007:
Apologies for taking so long to get back to this. I did try the name approach, and was not able to get it to work. I also read a comment somewhere in the codebase that indicated all widget names should be unique for something to do with styling or layout, so I'm not sure it's a good idea to pursue.

I'll hopefully have the attribute change you mentioned ready for review in the next few days.

djmattyg007:
I've updated my branch and force-pushed the changes as a single squashed commit. Thanks for being patient.

enaml/qt/qt_abstract_button.py
```diff
@@ -32,6 +33,8 @@ class QtAbstractButton(QtControl, ProxyAbstractButton):
     #: A reference to the widget created by the proxy
     widget = Typed(QAbstractButton)
 
+    _current_group = Typed(ButtonGroup)
```
MatthieuDartiailh: Please add a quick command describing what this is used for.

enaml/qt/qt_abstract_button.py
```diff
@@ -112,6 +117,12 @@ def set_icon_size(self, size):
         with self.geometry_guard():
             self.widget.setIconSize(QSize(*size))
 
+    def set_group(self, group):
+        if self._current_group:
```
MatthieuDartiailh: Same here a quick dosctring would be nice.

enaml/qt/qt_button_group.py
```diff
@@ -0,0 +1,28 @@
+from atom.api import Typed
+from enaml.widgets.button_group import ProxyButtonGroup
+
+from .QtWidgets import QButtonGroup
+
+from .qt_toolkit_object import QtToolkitObject
+
+
+class QtButtonGroup(QtToolkitObject, ProxyButtonGroup):
+    widget = Typed(QButtonGroup)
```
MatthieuDartiailh: Same here a docstring for the class and the attribute (feel free to copy paste from other classes for example the following is often used for widget #: A reference to the widget created by the proxy.)

enaml/qt/qt_button_group.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def create_widget(self):
```
MatthieuDartiailh: Same for all methods here.

enaml/widgets/abstract_button.py
```diff
@@ -53,6 +61,8 @@ class AbstractButton(Control):
     #: and indicates that an appropriate default should be used.
     icon_size = d_(Coerced(Size, (-1, -1)))
 
+    group = d_(ForwardTyped(ButtonGroup))
```
MatthieuDartiailh: Documentation

enaml/widgets/button_group.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def add_button(self, button):
+        raise NotImplementedError
+
+    def remove_button(self, button):
+        raise NotImplementedError
+
+
+class ButtonGroup(ToolkitObject):
+    exclusive = d_(Bool(True))
```
MatthieuDartiailh: Documentation

djmattyg007:
Thanks for the review. Unfortunately I'm travelling this week so I may not get to it until next week.

MatthieuDartiailh:
No problem. I want @sccolbert opinion before merging anyway and he is usually quite busy.

MatthieuDartiailh:
Ping @djmattyg007

MatthieuDartiailh:
@sccolbert as part the release for the `sync_time` attribute, it may worth also adding this. I will add the missing docstrings. Apart from those do you see any blocking issue with this ?

sccolbert:
LGTM modulo the missing documentation.

MatthieuDartiailh:
Thanks @sccolbert

MatthieuDartiailh:
@sccolbert while adding the missing docs I realized that `add_button` and `remove_button` on the declaration could lead to inconsistencies since they were not setting the group attribute of the button being added. As a consequence I removed them since they are explicitly needed, do you agree ?  
A side concern, currently we do not keep track of the buttons pertaining to a group on the group itself. This means that one cannot use the group to access the button part of that group. I feel it might be a useful feature in some cases but I would like your opinion.

MatthieuDartiailh:
@sccolbert I found a reasonable way to track which buttons belong to a group and added tests to check everything was going as expected. Ideally I would like to merge by the end of the month and do the promised release (if possible I would like to also include #358).

PR #349, opened 2019-04-18
Migrate to cppy and drop Python 2 support 

MatthieuDartiailh:
WIP dropping Python 2 support and using cppy for C extensions.


frmdstryr:
Good to see the loc going down. A few comments:

1.  I don't think this should be considered a "micro" release.  
2. From looking at the code I'm not sure why the migration to cppy is needed? It saves a few lines but adds another dependency. Why was it not used originally?

Still have not had time to try it out, will update when I do.

MatthieuDartiailh:
This will be a minor release the first commit is a left over and can be ignored. cppy unify the helpers used by all nucleic project and reduce duplication. I will also migrate to heap allocated types and two-phase module init (more modern c extensions) the way I already did in atom and kiwi in the similar PRs.

frmdstryr:
I gave it a try, haven't thoroughly tested but it appears to work fine.

MatthieuDartiailh:
Thanks @frmdstryr

Issue #357, opened 2019-07-25
Looper `loop_index` becomes invalid when items are reused


frmdstryr:
This has really confused me for a long time...

```
import random
from enaml.core.api import Looper
from enaml.widgets.api import Window, Container, Label, PushButton


enamldef Main(Window): window:
    attr data: list = ['a', 'b', 'c']
    Container:
        PushButton:
            text = 'Shuffle'
            clicked ::
                d = window.data[:]
                random.shuffle(d)
                window.data = d
        Looper:
            iterable << window.data
            Label:
                text << '{} {} {}'.format(loop_index, loop_item, data.index(loop_item))


```

Initially it shows

```
0 a 0
1 b 1
2 c 2
```

After shuffling it shows

```
2 c 0
0 a 1
1 b 2
```

I would expect `loop_index` to always match the index of the item within the iterable but that is not the case when items are reordered.

sccolbert:
Looper tries to be smart and caches the rendered widgets inside the loop and simply moves them to the new location when the looper re-renders. What this means is that you shouldn't modify the iterable in-place. Create a copy of it, and you should be good to go.

frmdstryr:
The problem actually came up in code that was using a new list each time but a few items in the list were in a different position (changing options for a select field with enaml-web).  

The only easy way to do it is to wrap the iterable with `enumerate` but that defeats the point of caching because even if the item moves the whole widget needs rerendered as the (index, item) tuple changes.  

The PR I pushed tries to retain the intent of caching when items are moved but still need to update some members that depend on the index.

MatthieuDartiailh:
@frmdstryr do you consider this is not worth fixing ? I was in vacation the last two weeks and missed this.

frmdstryr:
Not sure, I haven't run into issues with it often. It depends on how people use loop_index.  

If not fixed I think it should at least be documented. Feel free to reopen.

frmdstryr:
#358

PR #358, opened 2019-07-26
Do Not Merge: Potential change to Looper api

frmdstryr:
This is one potential way to address #357 but do not merge as this introduces a breaking change to the Looper api.  

It requires that  `loop_item` and `loop_index` be replaced with `loop.item` and `loop.index`.  Perhaps the original scope variables could be retained for compatibility and the docs updated to indicate which to use.

Any thoughts or other suggestions?


frmdstryr:
Easy enough to work around this

MatthieuDartiailh:
I agree it is a breaking change but it looks kind of neat too. I may revisit it at one point if I ever get to deprecate ContainerList in Atom and need to update the Looper.

frmdstryr:
Updated it to retain the original `loop_item` and `loop_index` scope variables so no old code is broken by the change.

MatthieuDartiailh:
Is it true that using the old attribute can lead to wrong values in the index ? And this why we need the new structure ? If so could you expand the docstring in that direction, since it is the best docs we have so far.

enaml/core/looper.py
```diff
@@ -34,20 +34,25 @@ class Looper(Pattern):
...
     will only create and destroy children for the items in the iterable
-    which have changed.
+    which have changed. When an item in the iterable is moved the
+    `loop.index` will be updated to reflect the new index.
 
-    The Looper works under the assumptions that the values stored in the
+    The Looper works under the assumption that the values stored in the
     iterable are unique.
 
+    The `loop_index` scope variable is depreciated in favor of `loop.index`
```
MatthieuDartiailh: `loop_item` should be deprecated too even though it is not broken.

frmdstryr: Updated

MatthieuDartiailh:
@frmdstryr could you update the release notes ? Try to make the deprecation quite visible since it is one of the few place users can learn about it.

releasenotes.rst
```diff
@@ -4,6 +4,14 @@ Enaml Release Notes
 0.11.0 - unreleased
 -------------------
 
+0.10.4 - unreleased
+-------------------
+
+- fix Looper's ``loop_index`` becoming invalid when items are reordered #357 via PR #358
+
+.. note:: Looper's ``loop_index`` and ``loop_item`` scope variables are now depreciated. When upgrading to 0.10.4 or newer all usage of ``loop_index`` and ``loop_item`` within a Looper should be replaced with ``loop.index`` and ``loop.item`` respectively. See #357 for details.
```
MatthieuDartiailh: depreciated -> deprecated

MatthieuDartiailh:
Thanks @frmdstryr

MatthieuDartiailh:
@frmdstryr the diff looks kind of off now. Could you rebase on master ?

Issue #351, opened 2019-07-05
SyntaxError when using `.` in attr types


frmdstryr:
Currently when attempting to set the type of an attr it cannot contain a `.` Ex:

```python
from enaml.widgets.api import *

import datetime
from datetime import date

enamldef Main(Window): window:
    attr date_t: date
    attr date_t2: datetime.date
```

It throws a syntax error

```python-traceback
Traceback (most recent call last):
  File "/Workspace/enaml/enaml/core/parser/base_parser.py", line 199, in parse
    return self.parser.parse(source, debug=0, lexer=lexer)
  File "/miniconda3/envs/enaml/lib/python3.7/site-packages/ply-3.11-py3.7.egg/ply/yacc.py", line 333, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/miniconda3/envs/enaml/lib/python3.7/site-packages/ply-3.11-py3.7.egg/ply/yacc.py", line 1201, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/miniconda3/envs/enaml/lib/python3.7/site-packages/ply-3.11-py3.7.egg/ply/yacc.py", line 192, in call_errorfunc
    r = errorfunc(token)
  File "/Workspace/enaml/enaml/core/parser/base_parser.py", line 3405, in p_error
    syntax_error(msg, FakeToken(lexer, t.lineno))
  File "/Workspace/enaml/enaml/core/parser/base_lexer.py", line 71, in syntax_error
    _parsing_error(SyntaxError, message, token)
  File "/Workspace/enaml/enaml/core/parser/base_lexer.py", line 55, in _parsing_error
    raise ParsingError(klass, message, info)
enaml.core.parser.base_lexer.ParsingError: ('invalid syntax', ('example.enaml', 9, 0, '    attr date_t2: datetime.date'))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/miniconda3/envs/enaml/bin/enaml-run", line 11, in <module>
    load_entry_point('enaml', 'console_scripts', 'enaml-run')()
  File "/Workspace/enaml/enaml/runner.py", line 45, in main
    ast = parse(enaml_code, filename=enaml_file)
  File "/Workspace/enaml/enaml/core/parser/__init__.py", line 48, in parse
    return _parser.parse(enaml_source, filename)
[...]

PR #359, opened 2019-08-18
Add support for . in attr types

frmdstryr:
I have not thoroughly tested this also I'm not sure if this covers the edge case(s) mentioned in https://github.com/nucleic/enaml/issues/351#issuecomment-508802683


MatthieuDartiailh:
@sccolbert did you remember the edge case you mentioned ?

enaml/core/compiler_common.py
```diff
@@ -259,6 +259,11 @@ def load_name(cg, name, local_names):
     """
     if name in local_names:
         cg.load_fast(name)
+    elif '.' in name:
+        parts = name.split('.')
+        load_name(cg, parts.pop(0), local_names)
+        while parts:
+            cg.load_attr(parts.pop(0))
```
sccolbert: This is the wrong place for this logic. The `name` argument is assumed to not have a dot. We should really make a new ast node to describe the dotted  type name, then do a more granular emit in the compiler.

Hiding the logic away in this function, while easy, is a bit too much magic for my taste.

MatthieuDartiailh:
This was not added upfront to avoid people being bitten by the fact that the type should only reference globally accessible objects since it is accessed during the 'class' definition and not its instantiation.
Currently people can already end up in trouble is they do this:
```
enamldef Main(Window):
    attr a
    attr b : a = 1
```

We will have to clearly document this behavior.

frmdstryr:
Thanks for the feedback. Instead of a separate ast node here would it be better to reuse the `dotted_name` and [ast_for_dotted_name](https://github.com/nucleic/enaml/blob/master/enaml/core/parser/base_parser.py#L51)?

sccolbert:
> Thanks for the feedback. Instead of a separate ast node here would it be better to reuse the `dotted_name` and [ast_for_dotted_name](https://github.com/nucleic/enaml/blob/master/enaml/core/parser/base_parser.py#L51)?

Yes. And then update the compiler to handle that AST node in the type position.

enaml/core/parser/base_parser.py
```diff
@@ -589,18 +593,21 @@ def p_storage_expr3(self, p):
...
         node.kind = kind
         node.name = p[2]
-        node.typename = p[4]
+        typename = ast_for_dotted_name(p[4])
+        typename.lineno = lineno
+        node.typename = typename
         node.expr = p[5]
         p[0] = node
 
+
```
MatthieuDartiailh: extra blank line

MatthieuDartiailh:
ping @sccolbert @frmdstryr

MatthieuDartiailh:
Ping @sccolbert

MatthieuDartiailh:
ping @sccolbert

sccolbert:
LGTM

MatthieuDartiailh:
Thanks @sccolbert I will merge on my next spare cycle.

PR #365, opened 2019-09-03
Fixing issue with validator fields

e-calder:
Fixing issue with validator fields, where clearing the error state does not restore the original stylesheet

### DIFF
--- a/enaml/qt/qt_field.py
+++ b/enaml/qt/qt_field.py
@@ -126,7 +126,8 @@ class QtField(QtControl, ProxyField):
         # A temporary hack until styles are implemented
         if self._guard & ERROR_FLAG:
             self._guard &= ~ERROR_FLAG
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
This is a partial fix but better than nothing. Could you make sure the lines are not more than 79 chars long (you need to wrap the comment) and add an entry in the releasenotes ? Also Would you mind opening an issue tracking the fact that we should integrate the Field validation with the existing StyleSheet machinery.

e-calder:
Thanks for looking. 

I have corrected the long comment such that the line lengths do not exceed 79 characters, and alsoo added an entry in the release notes.

The issue is created [here](https://github.com/nucleic/enaml/issues/366).

MatthieuDartiailh:
Great ! I keep delaying the next release but hopefully it will happen before the end of the month.

PR #378, opened 2019-10-11
Improve error reporting on invalid syntax for f-strings

### LINKED ISSUE #377: Backstring in f-string syntax error reporting wrong line
I should know better but the error message for using a backslash in an f-string in enaml is reporting the error at a line that doesn't make sense.

```python

    Container:
        attr path = r'C:\\fakepath\\SomeData.csv'
        PushButton:
            text = 'Set path'
            clicked ::
                # Backslash in f-string not allowed
                label.text = f'Import {path.split("\\")[-1]}'
        PushButton:
            text = 'Clear path'
            clicked :: label.text = ''
        Label: label:
            text = ''
```

It's reporting at line 13 but I'd expect it to be at the f-string line (10).  

```

  File "/home/user/miniconda3/envs/enaml3/bin/enaml-run", line 11, in <module>
    load_entry_point('enaml', 'console_scripts', 'enaml-run')()
  File "/home/user/Workspace/enaml/enaml/runner.py", line 45, in main
    ast = parse(enaml_code, filename=enaml_file)
  File "/home/user/Workspace/enaml/enaml/core/parser/__init__.py", line 48, in parse
    return _parser.parse(enaml_source, filename)
  File "/home/user/Workspace/enaml/enaml/core/parser/base_parser.py", line 201, in parse
    raise parse_error()
  File "tests/example.enaml", line 13
    clicked :: label.text = ''
    ^
SyntaxError: invalid syntax

```

It would be nice if the original error could be retained as well (if possible).
```python

path = r'C:\fakepath\SomeData.csv'
path
Out[2]: 'C:\\fakepath\\SomeData.csv'

x = f'Importing {path.split("\\")[-1]}...'
  File "<ipython-input-3-6ed1e7ff0f7d>", line 1
    x = f'Importing {path.split("\\")[-1]}...'
       ^
SyntaxError: f-string expression part cannot include a backslash
```

MatthieuDartiailh:
Closes #377 

This requires tests and release notes updates

### DIFF
--- a/enaml/core/parser/parser36.py
+++ b/enaml/core/parser/parser36.py
@@ -102,12 +102,16 @@ class Python36EnamlParser(Python35EnamlParser):
             m = ast.parse('(' + ' '.join([s for s, _ in p[1]]) + ')')
         except SyntaxError as e:
             err_msg = e.args[0]
+            # Identify the line at which the error occurred to get a more
+            # accurate line number
             for s, lineno in p[1]:
                 try:
                     m = ast.parse(s)
                 except SyntaxError:
                     break
 
+        # Avoid getting a triple nesting in the error report that does not
+        # bring anything relevant to the traceback.
         if err_msg:
             syntax_error(err_msg, FakeToken(p.lexer.lexer, lineno))
 
@@ -123,7 +127,6 @@ class Python36EnamlParser(Python35EnamlParser):
 
     def p_atom_string_list3(self, p):
         ''' atom_string_list : FSTRING '''
-        print(p.lineno(1))
         p[0] = [('f' + repr(p[1]), p.lineno(1))]
 
     def p_atom_string_list4(self, p):
--- a/enaml/core/parser/parser36.py
+++ b/enaml/core/parser/parser36.py
@@ -7,6 +7,7 @@
 #------------------------------------------------------------------------------
 import ast
 
+from .base_lexer import syntax_error
 from .lexer3 import Python36EnamlLexer
 from .base_parser import Store, ast_for_testlist, syntax_error, FakeToken
 from .parser35 import Python35EnamlParser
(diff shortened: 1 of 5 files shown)


enaml/core/parser/parser36.py
```diff
@@ -96,21 +97,35 @@ def p_annassign(self, p):
...
 
     def p_atom_string_list2(self, p):
         ''' atom_string_list : atom_string_list STRING '''
-        p[0] = p[1] + [repr(p[2])]
+        p[0] = p[1] + [(repr(p[2]), p.lineno(2))]
 
     def p_atom_string_list3(self, p):
         ''' atom_string_list : FSTRING '''
-        p[0] = ['f' + repr(p[1])]
+        print(p.lineno(1))
```
frmdstryr: print

enaml/core/parser/parser36.py
```diff
@@ -96,21 +97,35 @@ def p_annassign(self, p):
...
+            m = ast.parse('(' + ' '.join([s for s, _ in p[1]]) + ')')
+        except SyntaxError as e:
+            err_msg = e.args[0]
+            for s, lineno in p[1]:
+                try:
+                    m = ast.parse(s)
+                except SyntaxError:
+                    break
+
+        if err_msg:
```
frmdstryr: Is the separate `err_msg = ''` and if statement needed? Also it may be good to add a comment why the for loop is there.

MatthieuDartiailh: The if avoids Python giving a triple level exception (exception during error handling). Actually I could apply something similar in `parse` to avoid having multiple level at all.

MatthieuDartiailh:
Thanks for the review.

MatthieuDartiailh:
This is ready to merge from my point of view. Care to do a second review @frmdstryr ?

frmdstryr:
Works good. Also works for missing & empty brackets now too. Thanks!

Issue #380, opened 2019-10-29
Does `alias` need to be a reserved keyword?


frmdstryr:
Not a big deal but could `alias` be changed to not be a keyword and the `alias_expr` merged into `storage_expr` + validation? 

```python
from enaml.widgets.api import Window, Container, Label, PushButton

def build_query():
    import sqlalchemy as sa
    return str(sa.select().alias('blablablba'))

enamldef Main(Window): window:
    Label:
        text = build_query()
```

```python-traceback
Traceback (most recent call last):
  File "/home/user/miniconda3/envs/enaml3/bin/enaml-run", line 11, in <module>
    load_entry_point('enaml', 'console_scripts', 'enaml-run')()
  File "/home/user//enaml/enaml/runner.py", line 45, in main
    ast = parse(enaml_code, filename=enaml_file)
  File "/home/user//enaml/enaml/core/parser/__init__.py", line 48, in parse
    return _parser.parse(enaml_source, filename)
  File "/home/user//enaml/enaml/core/parser/base_parser.py", line 201, in parse
    raise parse_error()
  File "tests/example.enaml", line 7
    return str(sa.select().alias('blablablba'))
    ^
SyntaxError: invalid syntax
```

As a workaround I have to use getattr or push to a separate module... I can submit a PR to review if there are no objections.

sccolbert:
It should only be treated as keyword inside an `enamldef` block. If it’s not allowed outside such a block, that’s a bug we need to fix.

MatthieuDartiailh:
Currently none of the following enaml keywords behave this way: pragma, const, alias, enamldef, template. However attr and func does behave as described by being treated as NAME token. We could change that  for the others too. I fear it may make the grammar ambiguous though. Otherwise we could keep track of the scope in the lexer and turn special tokens to NAME if outside of the proper scope.
No matter the solution we should homogenize the behavior, I think.

MatthieuDartiailh:
Since alias and const follow similar rules to attr we can probably just add branches in the parser functions handling those rules. In that case there is no risk of ambiguity in the grammar. Not sure when I will have time to address this. I would like to finish the copy migration first. If anyone wants to tackle they are welcome.

MatthieuDartiailh:
In order to avoid repeating #381, I would like we agree on the principle of the solution to use. The way I see it we have two solutions:
[...]

PR #382, opened 2019-11-04
parser: allow alias and const not to be keywords

MatthieuDartiailh:
This implements the solution 2 outlines in #380. Except for the need to manually filter out possibly wrong nodes this looks reasonably clean to me.

Existing tests pass but we probably need some extra ones.

### DIFF
--- a/enaml/core/parser/base_parser.py
+++ b/enaml/core/parser/base_parser.py
@@ -439,8 +439,9 @@ class BaseEnamlParser(object):
     def p_enamldef_simple_item1(self, p):
         ''' enamldef_simple_item : binding
                                  | ex_binding
-                                 | alias_expr
-                                 | storage_expr '''
+                                 | storage_alias_const_expr '''
+        if isinstance(p[1], enaml_ast.ConstExpr):
+            syntax_error('invalid syntax', FakeToken(p.lexer.lexer, p[1].lineno))
         p[0] = p[1]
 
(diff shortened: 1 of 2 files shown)


enaml/core/parser/base_parser.py
```diff
@@ -495,125 +496,104 @@ def p_pragma_arg3(self, p):
...
+        # Storage expression
+        else:
+            self._validate_storage_expr(kind, lineno, p.lexer.lexer)
+            node = enaml_ast.StorageExpr()
+            node.lineno = lineno
+            node.kind = kind
+            node.name = p[2]
+            p[0] = node
+
+    def p_sstorage_alias_const_expr2(self, p):
```
frmdstryr: extra s in storage

MatthieuDartiailh: Good catch and proof we need more tests !!!

MatthieuDartiailh:
The one failure we have is under Python 2 and since I won't make a bugfix release for that it will be fine since we are dropping Python 2. I will try to merge the cppy PR soon to make that clear.

frmdstryr:
I'd rather have python 2 support than alias and const not being keywords.

MatthieuDartiailh:
I won't support Python 2 after january 2020 and I don't plan a release between now and then.

MatthieuDartiailh:
This is ready to be merged from my point of view and will be next week if nobody complains before then.

Issue #392, opened 2020-01-22
Support generator expressions in decl func blocks


frmdstryr:
Currently on `0.10.2` this gives a syntax error
```python
if any(s for s in some_list if s <= 10):
  # SyntaxError: generator expressions not allowed in a declarative function block
```

But wrapping it in a list works fine

```python
if any([s for s in some_list if s <= 10]):
   # Life is good
```

also commenting out https://github.com/nucleic/enaml/blob/master/enaml/core/parser/base_parser.py#L760 works fine.

Is this exclusion necessary?

MatthieuDartiailh:
I need to spend more time looking at the details of it but I believe this limitation comes from the need to run the generator expression in the proper scope to get have the right locals. I did quite a bit of work to get this to work for all comprehensions but never looked at generator expressions. 

Could you test it ? On straightforward way is by using `test_inner_functions.py` and adapting it to generator expr.

MatthieuDartiailh:
Also note that this limitation is not limited to declarative functions but also in event handling blocks.

MatthieuDartiailh:
Just for reference the rest of the discussion can be found in #393

PR #393, opened 2020-01-22
Allow gen expr in decl func def

frmdstryr:
For #392

### DIFF
--- a/tests/test_inner_functions.py
+++ b/tests/test_inner_functions.py
@@ -200,6 +200,17 @@ def test_set_comprehension_func(value):
     assert win.call()
 
 
+# Test that we can access the non local values and the closure variables
+@pytest.mark.parametrize('value', ['self', 'test'])
+def test_generator_expression_func(value):
+    """Test running a generator expression in a declarative function.
+
+    """
+    source = FUNCTION_TEMPLATE.format('any(%s for i in range(10))' % value)
+    win = compile_source(source, 'Main')()
+    assert win.call()
+
+
 def test_handling_nested_comprehension():
     """Test handling nested comprehensions.
 
--- a/enaml/core/parser/base_parser.py
+++ b/enaml/core/parser/base_parser.py
@@ -757,7 +757,6 @@ class BaseEnamlParser(object):
         ast.FunctionDef: 'function definition',
         ast.ClassDef: 'class definition',
         ast.Yield: 'yield statement',
-        ast.GeneratorExp: 'generator expressions',
     }
 
     def _validate_decl_funcdef(self, funcdef, lexer):


tests/test_inner_functions.py
```diff
@@ -200,6 +200,17 @@ def test_set_comprehension_func(value):
     assert win.call()
 
 
+# Test that we can access the non local values and the closure variables
+@pytest.mark.parametrize('value', ['self', 'test'])
+def test_generator_expression_func(value):
+    """Test running a generator expression in a declarative function.
+
+    """
+    source = FUNCTION_TEMPLATE.format('any(i for i in range(10))')
```
MatthieuDartiailh: You need to use value like in other tests otherwise you do not test access to non local values and closures.

MatthieuDartiailh:
It may be worth running something similar to the test but outside pytest to avoid mixing things. The Dynamicscope issue seems only related to how pytest try to report the issue. I agree it is worth fixing Dynamicscope though.

frmdstryr:
With #394 the error becomes

```
value = 'self'

    @pytest.mark.parametrize('value', ['self', 'test'])
    def test_generator_expression_func(value):
        """Test running a generator expression in a declarative function.
    
        """
        source = FUNCTION_TEMPLATE.format('any(%s for i in range(10))' % value)
        win = compile_source(source, 'Main')()
>       assert win.call()

tests/test_inner_functions.py:211: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
<test>:6: in call
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

.0 = <range_iterator object at 0x7ff9d111e810>

>   ???
E   NameError: name 'self' is not defined

```

It passes for `test`. The strange trace makes me think I something wrong with the DynamicScope.get.

MatthieuDartiailh:
Actually I also remembered another issue. Gen expr can also be simply returned from the function and for them to behave properly they would need to carry a reference to the Dynamicscope which is why they are not allowed. I don't feel confortable lifting that limitation actually. We could improve the documentation though.

frmdstryr:
Makes sense. So any "closure" that doesn't copy everything before leaving the scope is potentially an issue.

I guess it wouldn't make sense to force evaluation in this case? Eg if the funcdef contains a GenExpr add a decorator that checks if it returned a generator and if so convert it to a tuple?

MatthieuDartiailh:
Honestly I don't see the win in allowing GenExpr would bring us (knowing that list comprehension do work). And your idea would min the parser would need to change behavior based on the presence of the decorator which is not something I like (on top of the painful implementation).

I will close. Feel free to re-open if you feel the need.

PR #396, opened 2020-02-06
qt: fix q_dock_tab_widget for PyQt5

### LINKED ISSUE #395: Dock item alerting broken for tabbed items
Hi, 

Dock item alerting doesn't work for dock items that are tabbed inside a TabLayout:

**Repro:**
1. Run the dock_item_alerts example `enaml-run dock_item_alerts.enaml`

2. Drag the information dock item to the area with the Dummy items, so it ends up tabbed.

EnamlDrag

3. Now click the "Information" push button to alert it. You should see the error below:

```
  File "c:\users\[...]\enaml\qt\docking\q_dock_tab_widget.py", line 163, in _onAlerted
    self._snapAlertPixmaps(index, level)
  File "c:\users\[...]\enaml\qt\docking\q_dock_tab_widget.py", line 194, in _snapAlertPixmaps
    painter.initFrom(self)
AttributeError: 'QStylePainter' object has no attribute 'initFrom'
```

MatthieuDartiailh:
initFrom was removed and we need to do its job manually. Still need to manually set the pen and background but it already seems to be working.

Closes #395

### DIFF
--- a/enaml/qt/docking/q_dock_tab_widget.py
+++ b/enaml/qt/docking/q_dock_tab_widget.py
@@ -191,7 +191,7 @@ class QDockTabBar(QTabBar):
         normal = QPixmap(opt.rect.size())
         normal.fill(Qt.transparent)
         painter = QStylePainter(normal, self)
-        painter.setFont(self.font())
+        self.initPainter(painter)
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
This should be the right fix !

PR #402, opened 2020-04-01
CIs: switch to github actions

MatthieuDartiailh:
Keep appveyor for 32 bit windows


MatthieuDartiailh:
The issue on linux may be related to https://github.community/t5/GitHub-Actions/Seg-Faults-with-Python-in-Linux-while-Running-Pytest/td-p/39912 Won't touch for the time being.

Issue #401, opened 2020-03-31
NoFocus PushButton


AndiEcker:
As a total enaml newbie I struggle in declaring widgets (e.g. pushbuttons) that cannot receive the focus.

Is there a simple way to do this in the enamldef declaration, something like:

```
enamldef NoFocusButton(PushButton):
    proxy.widget.focusPolicy = QtCore.Qt.NoFocus
```

frmdstryr:
You can use the [activated](https://enaml.readthedocs.io/en/latest/api_ref/widgets/toolkit_object.html#enaml.widgets.toolkit_object.ToolkitObject.activated) event

eg

```python
enamldef NoFocusButton(PushButton):
    activated ::    
        self.proxy.widget.focusPolicy = QtCore.Qt.NoFocus
```

sccolbert:
Adding a `focus_policy` attribute would be a good addition to the base
`Widget` class:
https://github.com/nucleic/enaml/blob/master/enaml/widgets/widget.py#L134

AndiEcker:
thanks for your quick replies!!!

Unfortunately the workaround in changing the focusPolicy property of the qt
widget in the activated event did not work.

Will study further and try to implement a `focus_policy` attribute to the
widget class.


Am Di., 31. März 2020 um 19:15 Uhr schrieb S. Chris Colbert <
notifications@github.com>:

AndiEcker:
Are you sure @sccolbert that you want to have a `focus_policy` attribute
because you already had implemented it into enaml V 0.96
<https://github.com/nucleic/enaml/blob/e85a55ec957c361cb5ff5a4f781031dee0653ed1/releasenotes.rst#096---05082014>
but one version later it got removed.

Maybe you want that enaml users use instead your FocusTraversal feature,
something like:

```
enamldef NoFocusButton(PushButton):
    features = Feature.FocusTraversal

    previous_focus_child => (enaml_widget):
        return the_next_widget_that_can_receive_focus

    next_focus_child => (enaml_widget):
        return the_previous_widget_that_can_receive_focus
```

But when I set the FocusTraversal feature then I get an error at line 176
in
qt_widget.py (```if not widget.focusPolicy & Qt.TabFocus```), because
the focusPolicy attribute is actually a callable. FYI: in a previous enaml
version
this `if` expression was ```widget.focusPolicy*()* & Qt.TabFocus```)


¿Do you remember which problems you had with your implementation or why the
`focus_policy` attribute got removed in V 0.97?

[...]

PR #403, opened 2020-04-03
fixing bug in qt_widget.py/QtWidget.tab_focus_request().

AndiEcker:
- widget.focusPolicy is a function/callable that is missing the round brackets.
- usable for alternative workaround for issue #401.


MatthieuDartiailh:
Thanks could you add an entry in the release notes  and then I will merge ?

PR #404, opened 2020-04-05
fixed return value of QtWidget.tab_focus_request() and added release note entry

AndiEcker:
hope it will be PR 404

### DIFF
--- a/enaml/qt/qt_widget.py
+++ b/enaml/qt/qt_widget.py
@@ -180,7 +180,7 @@ class QtWidget(QtToolkitObject, ProxyWidget):
         if not widget.isVisibleTo(widget.window()):
             return False
         widget.setFocus(reason)
-        return False
+        return widget.hasFocus()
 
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
Thanks since this is the second bug in the same region. I will try to add some tests to avoid future regression before merging.

AndiEcker:
Thanks Matthieu - the new tests will be very helpful for me for to learn
how to also write some tests for the planned No-Focus implementation.

Am So., 5. Apr. 2020 um 20:30 Uhr schrieb Matthieu Dartiailh <
notifications@github.com>:

MatthieuDartiailh:
I added tests that currently fails on Travis (not sure why) however I was unable to reproduce your issue locally. Since the change seems reasonable though I will probably merge as soon as the Cis are fine.

MatthieuDartiailh:
Trying to kick Github actions

PR #409, opened 2020-04-08
prioritize enaml over py extension only within enaml.imports() context:

AndiEcker:
with these little changes we can have enaml
and py files within the same project or folder.

So e.g. having the files app.py and app.enaml
the code::

`
enaml.imports():
    from app import Main
`

will import `Main` from app.enaml and not
try to import it from app.py and then fail.


MatthieuDartiailh:
Close based on #410 discussion

Issue #411, opened 2020-04-11
Remove signaling/callableref/weakmethod


MatthieuDartiailh:
With the release of 0.11.0 I realized that signaling is broken. After a bit of investigation (and a lot of pain trying to fix it), I noticed that it has been true since 0.10.1 (basically it never worked on Python 3). 
Given that this was never reported before (and it has been more than two years now), I am in favor of simply removing signaling and the related callableref and weakmethod.

@sccolbert I remember you mentioning somebody having a use case for it. Could you ping them ? I can try to fix the situation but this is nasty part of C that I would be happy to just drop.

tstordyallison:
We do use this for our internal widgets (@ JPM). I would guess it was added by @sccolbert for this reason... 

We've already fixed up some ref counting issues with the version in 10.x (that we merged with our original fork in order to get Python 3 support, thanks for that!). So I wouldn't be too surprised if things aren't quite right. There aren't any tests in the repo for it and we picked it up on the CI for our widgets, not enaml itself.  

At this point - for our use cases - I'd be quite happy to pull in signaling/callableref/weakmethod into the C++ code base that sits along side our internal widgets and maintain it from there. We are on 2.7 + 3.7 in production today, with 2.7 going away towards the end of the year. I expect we'll start to move to 3.8 sometime thereafter and we'll fix up whatever we need to then. 

Less C++ to maintain needlessly in enaml.

MatthieuDartiailh:
Thanks for the feedback @tstordyallison !
Could you describe what is the added value of signaling in your case ? I would be happy to know because I have never come across a use case that would require it. Also since you have a use case for it, it may have value for others (once properly tested and documented) so it may make sense to keep it.
If you can provide the patch you applied to 0.10.x to get things to work I would be happy to apply it here (if this does not conflict with JP policies) and I can probably handle the cppy transition.

tstordyallison:
https://github.com/nucleic/enaml/blob/56163cb545a0f6b8a0b47eedf74bdf5896ac4c0b/enaml/src/weakmethod.cpp#L77

Dug out the PR - there was only actually one change (though it was in pythonhelpers.h in our fork) to set takeref=true. This seems to be missing on master too (e.g. `cppy::incref`).

As for the usage, I'll have more of a dig when I'm properly logged in tomorrow - IIRC it was to ease communications from the inside of a UI toolkit agnostic C++ widget -> enaml (a bit like Qt signals).

[...]

PR #412, opened 2020-04-11
Remove signaling and update cis

MatthieuDartiailh:


MatthieuDartiailh:
The failure first seen in Python 3.8 was the use of `METH_KEYWORDS` alone in `signaling.cpp`. Fixing that on 0.10.4 (before the cppy update), and adding the tests from https://github.com/nucleic/enaml/tree/signaling-testing uncover a lot more bugs including fairly cryptic ones where a python frame manage to replace a class ! This is real pain to track and is added motivation to simply remove it. 
I will leave this open for 2 more weeks and merge on May the 3rd, if anybody really need signaling please let yourself be known and if you can please try to investigate the issues.

MatthieuDartiailh:
Closed in favor of #416, thanks to the discussion in #411

Issue #414, opened 2020-04-16
Looper cannot refresh items when using a generator as the iterable


frmdstryr:
This took me several hours to debug in a large application... 

If the  `Looper` is given a generator as it's iterable, eg, `reversed` or `enumerate` and the items are refreshed multiple times without the `iterable` actually being updated (and thus reset) the iterable reaches the end and removes all items added to the looper!

This reproduces it:

```python
from enaml.core.api import Looper, Conditional
from enaml.widgets.api import Window, Container, Label, ObjectCombo

enamldef Main(Window): window:

    attr my_items = range(10)
    attr flag = 'a'
    attr show_labels: bool << flag == 'b' or flag == 'c'

    Container:
        ObjectCombo:
            items = ['a', 'b', 'c']
            selected := window.flag

        Conditional:
            condition << show_labels
            Label:
                text = 'Flag is B or C'

            # Nested loop on different attrs forces re-eval twice
            Conditional:
                condition << flag == 'c'
                Label:
                    text = 'Flag is C'
                Looper:
                    iterable << reversed(my_items)
                    Label:
                        text = str(loop_item)
```

It works fine if the combo is changed from `a` to `b` then `c` (it shows all labels from 0 to 9) but if you go straight from `a` to `c` the Looper removes all the items because the multiple levels of conditionals cause the looper to be refreshed multiple times without resetting the iterable back to the start.

I think to fix this iterable needs to be re-read from the expression engine every time it's refreshed.

MatthieuDartiailh:
This is quite tricky since even if we were re-reading each time you could still get bitten by doing `attr my_items = reversed(range(10))`. One thing we could do is test that the iterable is safe to iter multiple times using the following:
```python
it1 = iter(self.iterable)
it2 = iter(self.iterable)
if next(it1) != next(it2):
    raise ValueError("Iterable in Looper needs to be safe to iterate multiple times"
```
Testing this with range(10) works but fails with reversed(range(10))

Actually a better option could be to test that we get an iterable that is not an iterator using `collections.abc.Iterator`, that would be cleaner. Care to do a PR for that ?

PR #415, opened 2020-04-17
Coerce a Loopers iterable to an tuple if it is an iterator

frmdstryr:
For #414. 

Not sure if there's a better way to do this.

### DIFF
--- a/tests/core/test_looper.py
+++ b/tests/core/test_looper.py
@@ -122,6 +122,3 @@ def test_looper_iterable(enaml_qtbot, enaml_sleep):
 
     with pytest.raises(TypeError):
         looper.iterable = None
-
-
-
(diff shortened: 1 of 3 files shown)


frmdstryr:
Ooops, this doesn't handle non-iterable's properly...

tests/core/test_looper.py
```diff
@@ -104,3 +104,51 @@ def test_looper_refresh_iterator(enaml_qtbot, enaml_sleep):
     expected = ['{} {}'.format(*it) for it in enumerate(reversed(data))]
     assert [c.text for c in tester.container.children
             if isinstance(c, Label)] == expected
+
+
+@pytest.mark.skipif(not is_qt_available(), reason='Requires a Qt binding')
```
MatthieuDartiailh, line 109: I would like to see this test also test the case of passing an iterator since this is the case we are addressing. Also Do we really need Qt for testing this ? I feel like we should be able to validate the logic without a backend (but I may be missing something).

MatthieuDartiailh:
Also don't forget to update the release notes.

MatthieuDartiailh:
This looks all good. Thanks for the patch I will merge shortly.

PR #348, opened 2019-04-05
Add subscription block

frmdstryr:
Well that was easy.

PR #417, opened 2020-05-28
Declarative function reference counting issue

MatthieuDartiailh:
This fixes a bug identified in https://github.com/nucleic/enaml/pull/348, which boils down to a reference counting issue in declarative function that led to the deletion of the builtins dictionary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,6 @@ jobs:
           file: ./coverage.xml
           flags: unittests
           name: codecov-umbrella
-          yml: ./codecov.yml
           fail_ci_if_error: true
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
@@ -92,5 +91,4 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: unittests
           name: codecov-umbrella
-          yml: ./codecov.yml
           fail_ci_if_error: true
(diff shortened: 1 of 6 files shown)


MatthieuDartiailh:
@frmdstryr if you can confirm this fixes the issue it would be great. I am not sure why the double Looper was the only one showing that issue. Basically anything calling dynamic functions enough time can trigger that bug. Meaning it is really dangerous for long running application !

frmdstryr:
Yes it seems to be working here. 

I'll can also test it with another project tomorrow where I originally saw this.

Thank you!

frmdstryr:
> Meaning it is really dangerous for long running application !

Haha, yeah found that out when one of my enaml-web websites crashed after a month or so of running...

Are there any tools that you know of to watch for stuff like this? I know of https://valgrind.org/ but haven't used it enough to be able to do much with it.

MatthieuDartiailh:
Valgrind can be used to watch for memory leak from my understanding. Here the issue is spurious decref which internal to python. You could in the background of your website run a monitoring of the references and look for suspicious trend (constant increase, or decrease of supposedly long-lived objects). The gc module probably provides all the required tools to do something like that.

cg9999:
This fixed the crashing issues I had with inkcut. Thanks!

MatthieuDartiailh:
Thanks for the confirmation !

frmdstryr:
It's also no longer crashing in the app where I originally saw it. Nice work!

MatthieuDartiailh:
Thanks. I will do my best to cut the release and publish the wheels next week. And please do not hesitate to report that kind of issue. Hopefully that was the last one for the cppy update.

MatthieuDartiailh:
I cut and published the release and I strongly encourage all users to update ! I also audited all uses of function returning borrowed references and hopefully this was the last issue caused by the cppy migration.

MatthieuDartiailh:
conda package should show up later this week.

Issue #362, opened 2019-08-20
Python 3.8 support


MatthieuDartiailh:
This is a meta to track the changes associated with Python 3.8:

Required changes (meaning enaml won't run without them):
- bytecode changes:
  - MAP_ADD: the expected order on the TOS changed 
  - BREAK_LOOP, CONTINUE_LOOP, SETUP_LOOP, SETUP_EXCEPT have been removed and as a consequence the code_generator `for_loop` and `try_squash_raise` need to be updated.
  - ROT_FOUR, BEGIN_FINALLY, CALL_FINALLY, POP_FINALLY have been added which may impact `try_squash_raise`.
  - END_FINALLY and WITH_CLEANUP_START changed of behavior
  - CodeType signature changed

Those require to update wbyteplay and in particular the stack trace calculation and since we will be dropping Python 2 it may be the proper time to switch to https://github.com/vstinner/bytecode which a more modern interface to the Python bytecode.

Other changes (required for feature parity with normal Python):
- support for assignment expression (may cause grammar issue, and is tightly related to https://github.com/nucleic/enaml/pull/348)
- positional only argument (change to the parser but should not be a big problem, except it may mean the AST for function may have change which would imply more pervasive changes)
- f-strings now have a new formatting mode marked by '='
- continue is now allowed in finally clause (not sure we were actively preventing it on the parser side)
- the rules for unpacking in yield and return have changed.

Techjar:
That has become a significant problem as Python 3.8 is now in the main Arch Linux repository.

sccolbert:
@Techjar you're welcome to fix it, and make a PR.

MatthieuDartiailh:
Current master should run on Python 3.8 however the following points have not been addressed yet:
- support for assignment expression
- new rules for unpacking in yield and return

ShyPixie:
I've built cppy and enaml from git but I keep receiving a NameError about BREAK_LOOP not defined

MatthieuDartiailh:
Could you share your test file I will try to investigate ?

bburan:
Ran into this. Pretty sure it's a bad build uploaded to anaconda (at least for Linux). Here's a minimal working example:

    conda create -n enaml-test enaml
    conda activate enaml-test
    python test.py

Where `test.py` is:

    import enaml
    with enaml.imports():
        pass

Giving the traceback:

```
Traceback (most recent call last):
  File "test.py", line 2, in <module>
    with enaml.imports():
[...]

PR #422, opened 2020-08-23
Add support for missing Python 3.8 features

MatthieuDartiailh:
This covers the walrus operator := and the use of * in return and yield


MatthieuDartiailh:
Closes #362

PR #423, opened 2020-08-23
Add support for Pyside2 and avoid using deprecated API

MatthieuDartiailh:
Qt 5.15 started to mark as obsolete a number of feature related to Screen detection. We use the new alternative that remain compatible with qt 5.9+.


MatthieuDartiailh:
If anybody has access to a multiple monitor display, I would appreciate if somebody could test the behavior of popups when the parent is close to the edge.

Issue #428, opened 2020-11-15
Build without wheels (Linux)


adigitoleo:
Hi, not really an issue but more of a question. I'm trying to install from source, partly to learn and partly because I work mostly on Linux at the moment. I would prefer to use setuptools and avoid pip/wheels if possible (to manage everything through pacman). I've read [the docs](https://enaml.readthedocs.io/en/latest/get_started/installation.html#compiling-it-yourself-the-hard-way) and checked setup.py but I'm probably missing something. I have all of the dependencies:

```
python-pyqt5: 5.15.1-2
python-ply: 3.11-5
python-kiwisolver: 1.3.1-1
python-atom: 0.5.2-1
```

However, when I try to build with ``python setup.py build`` I get distutils errors, which I think are related to the C extensions:

```
WARNING: The wheel package is not available.
  ERROR: Command errored out with exit status 1:
   command: /usr/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-wheel-jjlg2tia/cppy/setup.py'"'"'; __file__='"'"'/tmp/pip-wheel-jjlg2tia/cppy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-zs557i_a
       cwd: /tmp/pip-wheel-jjlg2tia/cppy/
  Complete output (6 lines):
  usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: setup.py --help [cmd1 cmd2 ...]
     or: setup.py --help-commands
     or: setup.py cmd --help

  error: invalid command 'bdist_wheel'
  ----------------------------------------
  ERROR: Failed building wheel for cppy
ERROR: Failed to build one or more wheels
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/setuptools/installer.py", line 126, in fetch_build_egg
    subprocess.check_call(cmd)
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #429, opened 2020-11-24
Fix undocumented depencies

adigitoleo:
See #428

### DIFF
--- a/docs/source/get_started/installation.rst
+++ b/docs/source/get_started/installation.rst
@@ -93,9 +93,8 @@ Compiling it yourself: The Hard Way
 
 Building Enaml from scratch requires a few external dependencies. The
 sections below describe how to install Enaml and all of its dependencies from
-scratch, starting with the installation of a Python runtime. The instructions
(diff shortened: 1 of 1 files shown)


docs/source/get_started/installation.rst
```diff
@@ -117,8 +116,8 @@ instructions for adding the package to your system.
 .. _Setuptools Install: https://pypi.python.org/pypi/setuptools/1.1.6
 .. _Pip Install: https://pip.pypa.io/en/stable/installing/
 
-`Ply`_
-~~~~~~
+`Ply`_ and `Cppy`_
```
MatthieuDartiailh, line 119: Given how the rest of the document is organized, I am not sure why you chose to group those 2 together. I think it would be better for Cppy to have its own separate entries.

MatthieuDartiailh (changes requested):
One small change, the rest is good and I have no idea why the CIs coverage decided to drop but it is unrelated to this.

MatthieuDartiailh (approved):

PR #430, opened 2020-12-01
Allow to explicitly set Qt application name

adigitoleo:
The argument to QApplication is used to set the application name. On Linux (Xorg) this can be queried using the xprop utility. Specifically, this name will be used for the ``WM_CLASS`` property [1], which can be used by X-based window managers to manipulate window placement and app-specific behaviour (automatic floating, tiling, positioning, etc). The approach proposed here is used by Jupyter's QtConsole [2] for example.

[1] https://tronche.com/gui/x/icccm/sec-4.html#WM_CLASS
[2] jupyter/qtconsole#16

This is only for the Qt backend.. I'm not familiar with the other frameworks.

- [x] Modify enaml.qt.qt_application.QtApplication
- [x] Update releasenotes.rst
- [ ] Add test case

### DIFF
--- a/enaml/qt/qt_application.py
+++ b/enaml/qt/qt_application.py
@@ -29,6 +29,12 @@ class QtApplication(Application):
 
     def __init__(self,  appname=None):
         """ Initialize a QtApplication.
+        
+        Parameters
+        ----------
+        appname : str, optional
+            Explicit application name to use for setting the WM_CLASS attribute 
+            on Linux.
 
         """
         super(QtApplication, self).__init__()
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
The links provided sound convincing to me. Could you please update the changelog and then I will merge ?

adigitoleo:
OK I added this to releasenotes.rst, is that what you mean? Sorry if I misunderstood, I can fix it if it's incorrect.

MatthieuDartiailh:
That's good, however I am unlikely to release today, so please replace the date by unreleased. Also if you can add a Linux specific test for the new feature, it would be great.

Issue #432, opened 2021-01-05
Feature request: Grouping dock tabs into a dropdown


frmdstryr:
It'd be nice if there was a way to push tabs into a dropdown once it runs out of space, like this

![img](https://raw.githubusercontent.com/githubuser0xFFFF/Qt-Advanced-Docking-System/master/doc/tab_menu.gif)

From https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System#tab-menu-for-easy-handling-of-many-tabbed-dock-widgets, is that a Qt thing or did they add that in?

Edit: Looks like they just add it to the title bar, https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/blob/master/src/DockAreaTitleBar.cpp#L147

sccolbert:
It looks like that person has just cloned Enaml's docking framework in C++
(with no acknowledgement or credit given) and added some things to it.

PR #433, opened 2021-01-13
Add dock tab corner buttons

frmdstryr:
For #432 

This adds the QDockTitleBar as the corner widget to the QDockTabWidget with a new `tabs` button.

image

When the button is clicked it opens a context menu that lists all the tabs. When an action is clicked it sets that as the current tab.

image

It also keeps the close, maximize, and restore buttons.

The styling was messed up a little and I added some changes fix that. I'm not sure if that is specfic to my os.. Any help would be appreciated.


MatthieuDartiailh:
This looks good. I will test locally in the coming days and comment again. One small question though since I am not extremely familiar with that part of the codebase: it looks like you had to add code to handle maximize/restore and close in the custom corner widget which was previously handled by the framework. Is there a way to avoid this by inheriting from the right Qt class that is used as default corner widget ?

frmdstryr:
I don't think so... the tab widget seems to be treated like a layout similar to a splitter. I think the code was designed to only support maximizing a single dock item per area/window.  Maximizing a tab widget is more like maximizing a set of items.

Don't merge this yet as I think the way I'm restoring the layout needs changed.

Antother possible issue, I think this will not properly save and restore if maximized (I'm not using that so I didn't test it).

MatthieuDartiailh:
I did give this a quick run and indeed the save/restore functionality appear to be broken in your branch, otherwise this seems pretty neat. Let me know if you need help with anything.

frmdstryr:
Alright I think it's working properly now. The problem was when the tab widget was set to the maximize widget it was then no longer in the layout thus the layout saver wasn't including it. The "normal" way it works is the QDockContainer just passes the underlying QDockItem frame around (I think).

The workaround I used was to add a `QDockPlaceholder` that swaps itself with the tab widget and "saves it's place" in the dock layout when it's maximized and swaps back when restoring. Upon saving and building the layout it delegates to the widget it's holding the place for. 

There might be a cleaner way to do this, such as adding a `QDockWidgetContainer` which behaves similarly to the dock container. or something...

MatthieuDartiailh (approved):
This looks all good and works well locally. Ideally I would love to have some tests but I will merge as is for now.

Thanks for your contribution.

frmdstryr:
Thanks!

PR #438, opened 2021-02-05
Fix scrollarea example docs

adigitoleo:


MatthieuDartiailh (changes requested):
Since this file is auto-generated from the example file itself could you also fix the example file itself ? 

There is no need to regenerate all examples just for this, so changing both places will be sufficient. Thank you for your contribution.

MatthieuDartiailh (approved):

PR #439, opened 2021-02-06
Minor fixes to vim syntax highlighting

adigitoleo:
This fixes spurrious highlighting of characters immediately preceding
or following an enamlOperator. Also, the case of the notification
operator `::` followed by and EOL is now handled correctly. Lastly, the
`attr` keyword is highlighted using the Define group (like python
decorators, etc.)

### DIFF
--- a/tools/vim/syntax/enaml.vim
+++ b/tools/vim/syntax/enaml.vim
@@ -22,9 +22,10 @@ endif
 
 " Enaml extensions
 syn keyword enamlStatement      enamldef
+syn match enamlAttribute        "\%(\s*\)\zs\(attr\)\ze\%(\s\w\)"
 " FIXME: This captures the predefined operators, not any extensions that may be
 " added.
-syn match enamlOperator         "\%(\w\|\s\)\(::\|<<\|>>\|=\|:=\)\%(\w\|\s\)"
+syn match enamlOperator         "\%(\w\|\s\)\zs\(::\|<<\|>>\|=\|:=\)\ze\%(\w\|\s\|$\)"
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Could you share screenshots of the higlighting before and after the change ? I do not use vim personally and it would help me see if this patch makes vim highlighting closer to the other ones.

adigitoleo:
**Before**

Default neovim colorscheme:


My colorscheme:


---
**After**

Default neovim colorscheme:


My colorscheme:


---

This code doesn't work: the syntax at the second button is broken. That is to show the "spurrious" highlighting with the current syntax file, which is something I discovered  by accident. Mainly I'm just sharing this for the fix to the `::` operator which can be followed by a newline (the original regex didn't check for this).

I thought I would add a definition for `attr` while I am at it. Now I realise that `func` could also deserve special treatment... Or I can forgo these if it's inconsistent with the other editor tools.

adigitoleo:
In the new commit I have made `attr` and `func` use the same definition as `enamldef`, which looks consistent with python's `class` and `def` keywords, let me know what you think:

MatthieuDartiailh:
That is looking good. Note that alias could use the same treatment as attr and func. Just for you to know here is how it looks for me under VS Code which is based on the Sublime text syntax highlighting.
image

adigitoleo:
Thanks for your time and help, I've added `alias` as a special keyword also.

MatthieuDartiailh:
The improvements you made are already quite nice so I will merge those now. And you are obviously welcome to make another PR later. Thanks for your contribution !

PR #440, opened 2021-02-07
examples: prevent crash if executing a handler in the editor raises

### LINKED ISSUE #436: LiveEditor crash if view interaction creates exception
Using the following code in the view and pressing the button crashs the editor.

```
from enaml.widgets.api import *
from enaml.layout.api import *


enamldef DemoContainer(Container):
    attr model
    PushButton:
        clicked :: model.foo
```

MatthieuDartiailh:
Closes #436

Any error occurring will be redirected to the error pane, and the editor remain live. 

Note that the same approach should not be used in a general app since raising in a handler can leave the app in a corrupted state.

### DIFF
--- a/examples/applib/live_editor.enaml
+++ b/examples/applib/live_editor.enaml
@@ -12,6 +12,9 @@ stitched together to form a live Enaml code editor application.
 
 << autodoc-me >>
 """
+import sys
+import traceback
+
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
@Tillsten does that address your issue ? or would you like something more ?

PR #443, opened 2021-02-20
Fix python3.9 support

pbehnke:
python3.9 was broken due to missing import.

### DIFF
--- a/enaml/core/parser/base_parser.py
+++ b/enaml/core/parser/base_parser.py
@@ -2652,7 +2652,7 @@ class BaseEnamlParser(object):
         ''' subscript : ELLIPSIS '''
         p[0] = ast.Ellipsis()
 
-    def p_subcript2(self, p):
+    def p_subscript2(self, p):
         ''' subscript : test '''
         p[0] = ast.Index(value=p[1])
 
--- a/enaml/core/parser/parser39.py
+++ b/enaml/core/parser/parser39.py
@@ -7,8 +7,9 @@
 #------------------------------------------------------------------------------
 import ast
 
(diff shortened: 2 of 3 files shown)


MatthieuDartiailh:
Thanks for your contribution. I took the liberty to use the singleton Load defined in base_parser rather creating new instances, and added some comments since I could not remember the reason behind those changes. If you can check It does fix your issue it would be great.

pbehnke:
Hello,

Yes, I noticed the Load class in base_parser and initially tried that approach (https://github.com/nucleic/enaml/pull/443/commits/48c4b4748ab2d60b3dec69fb56896633fe28c29c), but the Load object defined there is an instance rather than a class and therefore is not callable. You were already calling Load() here, so went with importing it from the ast module since it seemed to have a minimal impact on the rest of the code. I've tested this with enaml-web and it seems to fix the issue, which was just this missing symbol (Load) when using python3.9. Would you prefer passing the same instance to each of the ast.Tuple() calls? I haven't tried this as it seemed inconstant with the original intention, but can try it out if you'd like.

Great work on this project, btw. I'm love the ideas you've presented here; the use of kiwisolver is brilliant.

MatthieuDartiailh:
I already made the change to use the singleton from base_parser but had not time to add tests, so if you could test on your end (or better write test) it would be great.

pbehnke:
Ah! My bad; I missed seeing your commit. I rebuilt with this new change and it appears to have fixed the issue! Thank you!

MatthieuDartiailh:
Thanks for checking ! I will merge then

PR #452, opened 2021-09-06
Add support for notebook tab styling

frmdstryr:
I couldn't figure out how to style notebook tabs? 

If there's already a way to do it please close this.


MatthieuDartiailh:
I will need to test locally (hopefully before the end of the week) but the changes look good to me and I don't think there is a different way to do this.

MatthieuDartiailh:
I took the liberty to add the ability to switch between the default style and the custom one to make the example more illustrative. Otherwise this is great. Can you just add a changelog entry and I will merge ?

PR #466, opened 2021-12-28
rtd: fix build configuration

MatthieuDartiailh:
Because I do not own the nucleic org I cannot get rtd feedback on PR but I enabled PR building on RTD, so we can at least check things are fine.


MatthieuDartiailh:
Actually RTD won't build PR without org level intergration... too bad

Issue #467, opened 2022-01-04
Crash when closing a floating dock window which has a maximised dock item inside it


e-calder:
Hey, 

I've noticed a crash related to maximised floating windows within a group... to reproduce:

1) Run `examples/widgets/dock_area.enaml`
2) Drag out 2 or more dock items so they are floating, and group them into their own floating dock window (terminology might not be accurate, please excuse)

![Enaml1](https://user-images.githubusercontent.com/37456137/148088432-36394431-9608-4d53-9846-5dfc4f4d3a6d.png)

3) **Maximise** one of these dock items within the dock window, then close the dock window via the outer red cross (circled).

![Enaml2](https://user-images.githubusercontent.com/37456137/148088559-d58e76cc-821c-4e8d-b3d8-35d45695ab47.png)

4) This results in a hard crash, killing the application: 

```
Traceback (most recent call last):
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\qt\q_deferred_caller.py", line 43, in customEvent
    event.callback(*event.args, **event.kwargs)
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\widgets\toolkit_object.py", line 169, in destroy
    super(ToolkitObject, self).destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\styling.py", line 298, in destroy
    super(Stylable, self).destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\core\declarative.py", line 128, in destroy
    super(Declarative, self).destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\core\object.py", line 114, in destroy
    child.destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\widgets\toolkit_object.py", line 169, in destroy
    super(ToolkitObject, self).destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\styling.py", line 298, in destroy
    super(Stylable, self).destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\core\declarative.py", line 128, in destroy
    super(Declarative, self).destroy()
  File "c:\users\ecalder\code\virtualenvs\bc-env\lib\site-packages\enaml\core\object.py", line 114, in destroy
[...]

PR #469, opened 2022-01-12
Dock area fixes

MatthieuDartiailh:
While investigating and fixing #467, I found a bunch of other issues that this PR addresses along with #467.

I do not have time to wind up the unit tests this deserves so I would appreciate if anybody can stress test this is a bit before merging. Each commit addresses a specific issue and can be reviewed independently. Hopefully commit messages are clear.

### DIFF
--- a/enaml/qt/docking/dock_manager.py
+++ b/enaml/qt/docking/dock_manager.py
@@ -339,6 +339,11 @@ class DockManager(Atom):
                 size = container.sizeHint()
                 geometries[container] = QRect(window.pos(), size)
             for container in containers:
+                # Return a maximized container to normal.
+                # This ensures the container is properly removed from all the
+                # layouts it is attached to and it ca be freed once closed.
+                if container.frame_state.item_is_maximized:
+                    container.showNormal()
                 if not container.close():
                     container.unplug()
                     container.float()
--- a/enaml/qt/docking/q_dock_container.py
+++ b/enaml/qt/docking/q_dock_container.py
@@ -10,6 +10,7 @@ from atom.api import Typed, Bool
 from enaml.qt.QtCore import Qt, QMargins, QPoint, QRect, QEvent, Signal
 from enaml.qt.QtGui import QIcon, QCursor
 from enaml.qt.QtWidgets import QApplication, QLayout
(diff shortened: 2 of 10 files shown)


bburan:
I played around with this using `psiexperiment` and it seems to work fine. However, I did notice some additional behavior quirks with the behavior of the dock area that I had not noticed before. These quirks are present in `main` as well, so not a side-effect of this PR. For example:

In the following screenshot, there is an "x" indicating that we could close the tab group. Yet, it is not active because at least one item in the tab group cannot be closed (e.g., `closable` attribute is set to False). When all items can be closed, it does work (and closes them all out).

image

I wonder if the "x" should not be shown if it will have no behavior?

MatthieuDartiailh:
Thanks for testing. We could either hide the x or close only all closable tabs and use a more specific tool tip. WDYT ?

Either way I will try to include the fix in this PR.

bburan:
@MatthieuDartiailh I would argue for hiding the x since it seems counter-intuitive to show the x if it does not actually have the capability to close out all tabs (regardless of closability of each tab and/or the tooltip shown).

MatthieuDartiailh:
Ok. I guess what I have in mind would make more sense as an entry in a context menu (I am thinking VS Code like, close all, close right, ...). But this I will leave for later.

sccolbert:
Another option would be to re-dock all the non-closable tabs to the main
dock area when the X is clicked on a floating window.

MatthieuDartiailh:
When closing a floating window non closable items or items whose closing is vetoed are re-floated and it works fine. The case of @bburan is specific to tabs floating or not and the handling of closing the whole notebook.

MatthieuDartiailh:
@bburan I updated the code to hide the close button on the tabs if no item can be closed. This is closer to the old behavior while being more correct. I feel it could also be more convenient when there are many tabs and a single one cannot be closed. Let me know if you find anything else. I will plan to merge next week if nobody complains in the meantime.

bburan:
That sounds great. Thank you.

MatthieuDartiailh:
At one point this will call for tests but I will merge as is for the time being.

PR #470, opened 2022-01-13
Add python fstring scintilla tokens

frmdstryr:
QsciLexerPython has tokens for fstrings now.


MatthieuDartiailh:
Look good. Please add an entry in the release notes.

PR #474, opened 2022-01-22
Pegen based new parser

MatthieuDartiailh:
This is a heavy WIP aiming at replacing the ply based parser by one based on pegen (i.e. the PEG parser generator used by CPython itself). 

By doing this I hope to:
- be able to support seamlessly and will little effort new Python syntax (i.e. match)
- have a clearer definition of the enaml syntax
- improve error reporting with proper line and offset tracking

Pegen has a 0.1 release to which I participated but will likely need some extra work around error reporting (the Python generated parser does not yet mimic the C one used by CPython).

Also note that this will imply to drop Python 3.7


sccolbert:
Hows the performance compared to the ply parser? (Im just curious).

MatthieuDartiailh:
I did not measure it yet but I plan to do so once I get the tests to pass.

frmdstryr:
Is anything blocking this? I've been working on getting KDevelop to support enaml and this parser sets the col_offsets.

frmdstryr:
I ran a quick benchmark using ipython's timeit on one of my projects. It was on 9 files loaded in in memory about 4k loc.

The existing parser is about 4-5 times faster. Both on python 3.10

```python
from enaml.core.parser import parse
# read files into dict of filename, source
# then benchmark with
%timeit [parse(s, f) for f, s in files.items()]
```
Existing parser (current master)

```py
296 ms ± 5.59 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
302 ms ± 18.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
290 ms ± 1.96 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Pegen parser

```py
1.31 s ± 5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.36 s ± 8.36 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.36 s ± 10.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

MatthieuDartiailh:
So this is definitively worse than I had hoped for. Could you try to measure separately the cost of lexing and of parsing ? Currently we use the builtin tokenize but I have changes (yet to be merged into pegen) that would allow to use a different tokenizer and it would be great to know if we can win anything on that front.

I think that a pegen based solution in pure Python will always be slower due to how the parser work but I hope we can make the change more tolerable. It also means I should look again into the option of precompiling enaml files during install.

frmdstryr:
I'm not sure if I'm doing it correctly but:

Existing parser
```python
from enaml.core.parser import _parser

def lex(source, filename):
    lexer = _parser.lexer(filename)
    lexer.input(source)
    return list(lexer.token_stream)
%timeit [lex(s, f) for f, s in files.items()]
```
101 ms ± 3.07 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)def

Using tokenize used in the pegen branch

```python
import tokenizer
from pegen tokenizer import Tokenizer
def lex(source):
    tok_stream = tokenize.generate_tokens(io.StringIO(source).readline)
    tokenizer = Tokenizer(tok_stream)
    try:
        while tokenizer.getnext():
            pass
    except StopIteration:  
        pass
    return tokenizer._tokens
%timeit [lex(s) for f, s in files.items()]
```
57.7 ms ± 1.04 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

So the lexing part seems to be faster.

MatthieuDartiailh:
It does not look obviously wrong to me. So once I am done with 3.11 base support I won't lack things to do when it comes to the parser.

frmdstryr:
Well with the pegen parser, KDevelop can now at least highlight some of the variables properly.  :) The current parser does not set the col offsets for any of the python expressions so everything highlights out of wack.
image

I would like to get it to somehow check all the attributes and understand the scoping but the ast does not really map directly to python and confuses it.

MatthieuDartiailh:
I am honestly impressed you get that much to work. Improving tooling around enaml files is something I would like to pursue but I have no idea when.

MatthieuDartiailh:
Another thing to check would be to regenerate the parser with the main branchbof pegen which includes better handling of error checks and some optimizations around unused variable. I wonder how much it influences performance.

sccolbert:
That's pretty cool!

This all has me wondering what a next-gen Enaml would look like if we owned
the whole stack all the way down to rendering.

MatthieuDartiailh:
One project that reminds me of enaml (even though I did not dive deep enough) is https://slint-ui.com/releases/0.3.1/docs/rust/slint/docs/langref/index.html

frmdstryr:
The code is in https://github.com/frmdstryr/kdev-python/tree/enaml-support . Their parser originally used the c-api but since that was removed in 3.10 I changed it to use the python api and was able to just add the enaml stuff on top.

MatthieuDartiailh:
@frmdstryr I will try to merge this by the end of the week so as to unlock work on 3.11 support. I will open issues for the remaining pain points (speed, compilation to enamlc at install time, etc) since my current bandwidth won't allow me to get to them in a reasonable time.

frmdstryr:
Thanks Matthieu!

MatthieuDartiailh:
With the last fix all exopy tests pass and since our tests pass too I will consider this good enough to merge.

PR #476, opened 2022-02-07
core: avoid generating empty block when trimming bytecode blocks

MatthieuDartiailh:
This is missing a test but should otherwise be good.


MatthieuDartiailh:
The appveyor failure is unrelated and not something I have time to investigate ATM.

Issue #473, opened 2022-01-21
Mechanism to provide custom size hints


MatthieuDartiailh:
In some situation the toolkit provided size hint may be less than ideal. Matplotlib widgets are a perfect example of this as discussed on the mailing list: https://groups.google.com/d/msgid/enaml/CAPc_zRWYW0jUMP-m2YrYNmCihryV4Lgr_VUvJAQjQWM1gPU%2BGA%40mail.gmail.com?utm_medium=email&utm_source=footer

Providing a declarative way to override the toolkit size hint would be nice since the procedure to ignore the existing size hint based constraints and generating new ones is not trivial.

@sccolbert any take on the API we may want for this (attribute vs declarative method) ?

@bburan once we settle on an API I will try to implement the solution but since I have other enaml work that has been delayed a bit it may take some time. Feel free to make a PR yourself if you are so inclined.

bburan:
@MatthieuDartiailh Happy to take a stab at implementing once we come up with the API. The `MainWindow` class does have an `initial_size` attribute which seems along the lines of what one might expect (e.g., add a `size_hint` option). What's the basis for deciding between an attribute vs. method in this case?

sccolbert:
As far as attribute vs. method, making it a method would denote something
needs to be computed. Qt makes it a method, but you have access to
information like text measurement from which to compute a preferred size.
You won't have that at the Enaml level. Everything you compute would need
to be derived from information provided entirely from the Enaml side
(exposing the toolkit-provided size hint is probably impossible to make
work for a variety of reasons).

My first instinct would be to make it an attribute, but I could probably
drum up a use case for making it a method.

MatthieuDartiailh:
Going for a method would allow to retrieve the toolkit hint and then adapt it. It would also allow to adjust the guess based on the parent content. In the case discussed in the mailing list, one could retrieve the number of plots in the grid to provide a better hint.

Also with an attribute I could see people expecting that updating the attribute would update the layout.

So personally I favor a method whose default implementation retrieves the toolkit hint.

bburan:
A method does add flexibility and if all one wants to do is provide a size with no calculations, it's still fairly readable code, e.g.:
```
MPLContainer:
    size_hint => ():
        return (800, 400)
```
as opposed to:
```
MPLContainer:
    size_hint = (800, 400)
```

[...]

PR #478, opened 2022-02-17
Add example for overridding size hints

bburan:
This addresses PR #473

### DIFF
--- a/examples/layout/advanced/mpl_canvas_size.enaml
+++ b/examples/layout/advanced/mpl_canvas_size.enaml
@@ -1,0 +1,76 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+#------------------------------------------------------------------------------
+""" An example of how to override tolkit-supplied size hints on your widget.
+
+
+The toolkit-supplied size hint for MPLCanvas is 480 x 640 pixels, which
+prevents you from sizing the canvas smaller. This can be a problem when you are
+presenting a large grid of Matplotlib figures.
+
+This example demonstrates how to override the tookit-supplied size hint for
+MPLCanvas. By setting resist_width, resist_height, hug_height, and hug_width to
+'ignore', you can resize the figure almost freely in any direction.
+
+This serves to demonstrate how the resist, hug and limit constraints interact
+to control the size of a widget.
+
+<< autodoc-me >>
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from enaml.widgets.api import (Form, Label, MPLCanvas, ObjectCombo, VGroup,
+                               Window)
+
+
+def make_figure():
+    figure, axes = plt.subplots()
+    x = np.arange(1000) / 1000
+    y = np.sin(2 * np.pi * 5 * x)
+    axes.plot(x, y)
+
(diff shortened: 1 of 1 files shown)


examples/layout/advanced/mpl_canvas_size.enaml
```diff
@@ -0,0 +1,76 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team.
```
MatthieuDartiailh, line 2: Please update the copyright year to 2022

examples/layout/advanced/mpl_canvas_size.enaml
```diff
@@ -0,0 +1,76 @@
...
+def make_figure():
+    figure, axes = plt.subplots()
+    x = np.arange(1000) / 1000
+    y = np.sin(2 * np.pi * 5 * x)
+    axes.plot(x, y)
+
+    # Call the `tight_layout` method anytime the user resizes the main window.
+    # This ensures that the axes titles and axes ticks are adjusted so they
+    # remain visible even at small sizes..
+    figure.canvas.mpl_connect('resize_event', lambda e: figure.tight_layout())
```
MatthieuDartiailh, line 41: You could enable constrained layout on your figure, plt.subplots(constrained_layout=True). It uses kiwisolver and give more consistent layout IMO. However I don't know if it will auto-refresh, or how to refresh it.

examples/layout/advanced/mpl_canvas_size.enaml
```diff
@@ -0,0 +1,76 @@
...
+    return figure
+
+
+enamldef Main(Window):
+
+    Form:
+        Label:
```
MatthieuDartiailh, line 48: May be worth detailing what resist_, hug_, limit_ impact each. So that people don't need to go find it in the source for themselves which is not exactly trivial.

MatthieuDartiailh (commented):
Looks good but I suggested some minor improvements. Also could you run the example_doc_generator.py found in docs/source/example and commit just the results relevant to this example.

MatthieuDartiailh:
You can ignore the 3.7 failures, I will drop support for it in the next version (done in a pending PR)

examples/layout/advanced/mpl_canvas_size.enaml
bburan: I didn't realize that constrained layout was based on kiwisolver. That's great. It also seems to auto-refresh so that's a nice bonus.

examples/layout/advanced/mpl_canvas_size.enaml
bburan: Done.

examples/layout/advanced/mpl_canvas_size.enaml
bburan: Done.

bburan:
I ran the example generator script but only committed the screenshot for the new example. I use Ubuntu, so the screenshot reflects the window decorations for that platform. Is that OK? I don't have access to a Mac OS computer.

MatthieuDartiailh:
I do not have a Mac anymore so even if I were to regenerate the examples the look would change so don't worry about that. Note however that you need to include the new example in the toctree of the documentation. That will fix the documentation CI.

Otherwise the changes looks great and I am happy to know that constrained layout worked well. Anytime I need to release a new version of kiwi, I sweat a bit since if I mess up the feedback is rather fast !

MatthieuDartiailh:
Thanks ! since the CIs failure are unrelated I will merge.

PR #479, opened 2022-02-18
Add new declarative @observe which ignores 'create' events by default

frmdstryr:
In order to avoid unnecessary calls to to `_update_proxy` (which by default only handles the `update` event type anyways), this change adds a new `observe` decorator in `enaml.core.declarative` that updates the default `change_types` flag to ignore `create` events. The widget declarations were updated to use the new import instead of the one from atom.api.  

The static observer added by the declarative metaclass will also ignore `created` events.

Any 3rd-party code using the original atom.api for custom members will be unchanged unless it relied on a create event from a superclass member. 

If someone needs the create event, either use the `observe` decorator from the atom.api or pass in the `change_types` flags for the needed. If the member had `@observe` on a super class, you can re-bind it with new flags to change the events triggerd.

### DIFF
--- a/enaml/core/declarative_meta.py
+++ b/enaml/core/declarative_meta.py
@@ -5,7 +5,11 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
-from atom.api import Atom, AtomMeta, DefaultValue, Member, Typed
+from atom.api import Atom, AtomMeta, ChangeType, DefaultValue, Member, Typed
+
+
+#: The Declarative engine ignores the create event type
+D_CHANGE_TYPES = ChangeType.ANY & ~ChangeType.CREATE
 
 
 class DeclarativeDefaultHandler(Atom):
@@ -88,7 +92,7 @@ def patch_d_member(member):
         new_mode = DefaultValue.CallObject_ObjectName
         member.set_default_value_mode(new_mode, handler)
     if metadata['d_readable']:
-        member.add_static_observer(declarative_change_handler)
+        member.add_static_observer(declarative_change_handler, D_CHANGE_TYPES)
 
 
 class DeclarativeMeta(AtomMeta):
--- a/enaml/core/declarative_meta.py
+++ b/enaml/core/declarative_meta.py
@@ -8,7 +8,9 @@
 from atom.api import Atom, AtomMeta, ChangeType, DefaultValue, Member, Typed
 
 # The declarative engine only updates for these types of changes
-D_CHANGE_TYPES = ChangeType.UPDATE | ChangeType.PROPERTY | ChangeType.EVENT
+D_CHANGE_TYPES = (
+    ChangeType.UPDATE | ChangeType.PROPERTY | ChangeType.EVENT | ChangeType.DELETE
+)
 
 
 class DeclarativeDefaultHandler(Atom):
--- a/enaml/core/declarative_meta.py
+++ b/enaml/core/declarative_meta.py
@@ -7,9 +7,8 @@
 #------------------------------------------------------------------------------
 from atom.api import Atom, AtomMeta, ChangeType, DefaultValue, Member, Typed
 
-
-#: The Declarative engine ignores the create event type
-D_CHANGE_TYPES = ChangeType.ANY & ~ChangeType.CREATE
+# The declarative engine only updates for these types of changes
+D_CHANGE_TYPES = ChangeType.UPDATE | ChangeType.PROPERTY | ChangeType.EVENT
 
 
 class DeclarativeDefaultHandler(Atom):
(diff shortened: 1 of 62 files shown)


MatthieuDartiailh:
This looks really nice !

enaml/core/declarative_meta.py
```diff
@@ -88,7 +92,7 @@ def patch_d_member(member):
         new_mode = DefaultValue.CallObject_ObjectName
         member.set_default_value_mode(new_mode, handler)
     if metadata['d_readable']:
-        member.add_static_observer(declarative_change_handler)
+        member.add_static_observer(declarative_change_handler, D_CHANGE_TYPES)
```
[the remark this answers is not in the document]
MatthieuDartiailh: In particular, I am wondering if the absence of container type change does not explain some issue I had in the past trying to implement a ListView.
@sccolbert can you tell if container is missing by design or is it an oversight ? And same qiestion for delete.

MatthieuDartiailh:
The 3.7 fails are expected we are dropping support for it bit the CI update is a pending PR.

enaml/core/declarative_meta.py
```diff
@@ -5,7 +5,10 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
-from atom.api import Atom, AtomMeta, DefaultValue, Member, Typed
+from atom.api import Atom, AtomMeta, ChangeType, DefaultValue, Member, Typed
+
+# The declarative engine only updates for these types of changes
+D_CHANGE_TYPES = ChangeType.UPDATE | ChangeType.PROPERTY | ChangeType.EVENT
```
MatthieuDartiailh, line 11: I feel like the previous version was more sensible, but let see if @sccolbert has an opinion.

frmdstryr: After thinking about it a little, the only way it could send the wrong change types would be if someone re-adds the `declarative_change_handler` with different flags.

Since that is unlikley to happen unintentially, I figured its safe to do.

I added a new set of flags for the observe decorator so that still keeps everything except create.

MatthieuDartiailh: I wanted to write and test a minimal demonstration, but since I can not find the time, let me try to demonstrate why I believe delete and container should be on that list.

```python
enamldef A(Container): a:
    attr t: str = "test"
    
    Field:
        test := a.t

    PushButton:
        text = "Reset"
        clicked:: del a.t
```

I often use the ability of atom to delete an attribute to restore it to its default. If the declarative engine ignores delete, it won't update the field here, which I believe is wrong.

WDYT ? I think a similar argument holds for container.

frmdstryr: They weren't used before unless I'm missing something: 
https://github.com/nucleic/enaml/blob/main/enaml/core/declarative_meta.py#L644

I think it'd be best to keep it the same then try enabling those in a separate PR.

MatthieuDartiailh: We could also check both project history. It is possible this bit predates delete and container notifications.

frmdstryr: It looks like delete was there 6 months before the property one was added. 

I guess I can just try it and see if there's any noticeable difference.

MatthieuDartiailh: Please do. Short of a direct answer from @sccolbert, this is the best path forward since we both agree it would make sense for it to be present.

MatthieuDartiailh:
I managed to fix the CIs though so could you rebase so that at least this is green ?

enaml/core/declarative_meta.py
sccolbert: I don't think I ever enabled container event listening, because I was never really happy with my implementation of reactive containers. For example, someone calling `.append` in a loop on a container list would end in disaster if the events were fired off synchronously. 

If I remember correctly, I was intending to add a dispatch mechanism which would batch the changes into a single event, then fire off that event when control returned to the event loop. But that would be a pretty big can of worms, probably best left to a major update/revision where we can streamline a bunch of stuff by not trying to be toolkit backend agnostic.

MatthieuDartiailh: I understand this concern. So I propose that for this PR we enable delete, leave container out and add somewhere in tye documentation the special behavior of enaml with respect to create and container.

Reworking notification should first happen in atom and while this is something I have been thinking about for years, I do not seem to be able to find the time to make any progress on it.

MatthieuDartiailh:
Actually for the docs in their current state I do not see where to add the information. It is a bit advanced for the getting started. it may have to wait for a doc overhaul...

MatthieuDartiailh:
@frmdstryr let me know your opinion regarding the documentation but with an entry in the release notes I believe this is good to go.

MatthieuDartiailh:
I think so, too, but this is currently empty. If you are inspired, feel free to get the section started.

releasenotes.rst
```diff
@@ -9,6 +9,7 @@ Dates are written as DD/MM/YYYY
 - fixes several bugs in corner cases of the Qt dock area PR #469
 - add python fstring scintilla tokens PR #470
 - address PyQt deprecation of accepting float values for pixel dimensions PR #471 and #481
+- add new declarative @observe which ignores 'create' events by default PR #479
```
MatthieuDartiailh, line 12:
```suggestion
- add new declarative @observe which ignores 'create' and 'container' events by default PR #479
- use atom builtin filtering mechanism to avoid refreshing the declarative engine on 
  'create' and 'container' event PR # 479
```

PR #471, opened 2022-01-14
Ensure pixel dimensions remain integer

buranconsult:
PyQt 5.15.6 may have deprecated some overloaded methods and methods and constructors no longer appear to accept floats. This fixes #472.

PR #481, opened 2022-03-04
Ensure some more QPoint values are ints

frmdstryr:


MatthieuDartiailh:
Looks good. Did you get a chance to audit other QPoint uses across the project ? If not I will try to do it. I would like for this PR to be the last regarding this issue.

frmdstryr:
After getting an error from the first change I did a search for QPoint and these are the only ones I found.  It looks like most were done in #471

MatthieuDartiailh:
Thanks for checking. Could add a release note entry (maybe just update the line of #471) ? I will merge ASAP.

MatthieuDartiailh:
Thanks for checking. Could add a release note entry (maybe just update the line of #471) ? I will merge ASAP.

PR #475, opened 2022-01-24
Qt6 support

MatthieuDartiailh:
WIP on adding support for Qt6

PR #484, opened 2022-03-27
Update scintilla for Qt6

frmdstryr:
Builds on #475. It seems like PyQt6 changes how enums values are exposed.  

One fix in q_icon_widget that may need addressed https://doc-snapshots.qt.io/qt6-dev/qlayout.html#getContentsMargins seems to have moved from QWidget.

### DIFF
--- a/enaml/qt/Qsci.py
+++ b/enaml/qt/Qsci.py
@@ -1,0 +1,19 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2022, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+#------------------------------------------------------------------------------
+from . import QT_API, PYSIDE2_API, PYQT5_API, PYQT6_API
+
+if QT_API in PYSIDE2_API:
+    msg = 'the Qt Scintilla widget is only available when using PyQt'
+    raise ImportError(msg)
+
+if QT_API in PYQT6_API:
+    from PyQt6.Qsci import *
+elif QT_API in PYQT5_API:
+    from PyQt5.Qsci import *
+else:
+    from QScintilla import *
--- a/enaml/qt/Qsci.py
+++ b/enaml/qt/Qsci.py
@@ -5,15 +5,12 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
-from . import QT_API, PYSIDE2_API, PYQT5_API, PYQT6_API
-
-if QT_API in PYSIDE2_API:
-    msg = 'the Qt Scintilla widget is only available when using PyQt'
-    raise ImportError(msg)
+from . import QT_API, PYQT5_API, PYQT6_API
 
 if QT_API in PYQT6_API:
     from PyQt6.Qsci import *
 elif QT_API in PYQT5_API:
     from PyQt5.Qsci import *
 else:
-    from QScintilla import *
+    msg = 'the Qt Scintilla widget is only available when using PyQt5 or PyQt6'
+    raise ImportError(msg)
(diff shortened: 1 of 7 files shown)


enaml/qt/Qsci.py
```diff
@@ -0,0 +1,19 @@
...
+
+if QT_API in PYSIDE2_API:
+    msg = 'the Qt Scintilla widget is only available when using PyQt'
+    raise ImportError(msg)
+
+if QT_API in PYQT6_API:
+    from PyQt6.Qsci import *
+elif QT_API in PYQT5_API:
+    from PyQt5.Qsci import *
+else:
```
MatthieuDartiailh, line 18: This dates from PyQt4 and can be removed. Also what is the situation with PySide6 ?

enaml/qt/docking/q_icon_widget.py
```diff
@@ -93,6 +94,8 @@ def minimumSizeHint(self):
         size = self._icon_size
         if not size.isValid():
             size = QSize(16, 16)
+        if IS_QT6:
```
MatthieuDartiailh, line 97: Could you add a comment on this one ?

MatthieuDartiailh (changes requested):
Since this was not caught in the CIs anything we could change to avoid future regressions ?

enaml/qt/docking/q_icon_widget.py
frmdstryr: The function getContentsMargins is obsolete and https://doc.qt.io/qt-5/qwidget-obsolete.html 
it seems it has been removed.

enaml/qt/Qsci.py
frmdstryr: I think pyside avoids it due to the GPL.

enaml/qt/docking/q_icon_widget.py
frmdstryr: Updated to use `contentsMargins()` now

enaml/qt/docking/q_icon_widget.py
```diff
@@ -93,8 +93,8 @@ def minimumSizeHint(self):
         size = self._icon_size
         if not size.isValid():
             size = QSize(16, 16)
-        left, top, right, bottom = self.getContentsMargins()
-        return size + QSize(left + right, top + bottom)
+        m = self.contentsMargins()
```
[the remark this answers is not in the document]
frmdstryr: A [search](https://github.com/nucleic/enaml/search?q=getContentsMargins) shows uses in `q_flow_layout.py` and `q_single_widget_layout.py` Both subclass QLayout which still has the method in Qt6.  Looks like you already did the label so I think that's all.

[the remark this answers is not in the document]
MatthieuDartiailh: Thanks for double checking

MatthieuDartiailh:
Looks great, I am just bothered by the fact that running the example in CI did not catch any of those issues. How did you catch them ?

frmdstryr:
These were found by running DeclaraCAD. 

It doesn't look like CI is running the live editor on Qt6?

MatthieuDartiailh:
Indeed looks like I messed up the CI config. Could you replace the following snippet by the next one in ci.yml ?

```
 - name: Install QScintilla
        if: matrix.python-version == '3.9' && matrix.qt-binding == 'pyqt'
        run: |
          pip install Cython QScintilla
```

```
 - name: Install QScintilla (Qt5)
        if: matrix.python-version == '3.9' && matrix.qt-binding == 'pyqt' && matrix.qt-version== 5
        run: |
          pip install QScintilla
 - name: Install QScintilla (Qt6)
        if: matrix.python-version == '3.9' && matrix.qt-binding == 'pyqt' && matrix.qt-version== 6
        run: |
          pip install PyQt6-QScintilla
```

MatthieuDartiailh:
Thanks. I will merge since otherwise the CIs won't kick in.

frmdstryr:
Qt6 seems to run a bit smoother, at least on linux & wayland. Thanks for doing the updates!

PR #486, opened 2022-04-01
Enaml qt6 fixes (for deprecated methods)

bburan:
I discovered all of these issues when testing with the desktop example. 

I have started testing against my application (psiexperiment) and may have more to report later.

### Examples to check
- [ ] aliases
- [ ] applib
- [ ] dynamic
- [ ] functions
- [ ] layout
- [ ] stdlib
- [ ] styling
- [ ] templates
- [ ] tutorial
- [x] widgets
- [ ] workbench

### Broken examples
- [ ] `drag_and_drop.enaml` (drag and drop broken)
- [ ] `file_dialog.enaml` (browse button broken)
- [ ] `flow_area.enaml` (changing flow layout)
- [ ] `image_view.enaml` (set to None)
- [ ] `ipython_console.enaml` (possibly just need to install qtconsole),
- [ ] `vtk_canvas.enaml` (most likely need to install vtk)
- [ ] `web_view.enaml`

### DIFF
--- a/enaml/qt/docking/__init__.py
+++ b/enaml/qt/docking/__init__.py
@@ -8,7 +8,9 @@
 from .. import QT_VERSION
 
 
-def hover_event_pos(event):
-    if QT_VERSION[0] == "6":
+if QT_VERSION[0] == "6":
+    def hover_event_pos(event):
         return event.position().toPoint()
-    return event.pos()
+else:
+    def hover_event_pos(event):
+        return event.pos()
--- a/enaml/qt/docking/__init__.py
+++ b/enaml/qt/docking/__init__.py
@@ -1,7 +1,14 @@
 #------------------------------------------------------------------------------
-# Copyright (c) 2013-2017, Nucleic Development Team.
+# Copyright (c) 2013-2022, Nucleic Development Team.
 #
 # Distributed under the terms of the Modified BSD License.
 #
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
+from .. import QT_VERSION
+
+
+def hover_event_pos(event):
+    if QT_VERSION[0] == "6":
+        return event.position().toPoint()
+    return event.pos()
--- a/enaml/qt/compat.py
+++ b/enaml/qt/compat.py
@@ -9,14 +9,18 @@ from qtpy.QtCore import QEvent
 
 from . import QT_API, PYQT6_API, QT_VERSION
 
-def coerce_to_qevent_type(event_value):
-    if QT_API in PYQT6_API:
-        return event_value
-    return QEvent.Type(event_value)
 
-
-def global_pos_from_event(event):
-    if QT_VERSION[0] == "6":
+if QT_VERSION[0] == '6':
+    def global_pos_from_event(event):
         return event.globalPosition().toPoint()
-    else:
+else:
+    def global_pos_from_event(event):
         return event.globalPos()
+
+
+if QT_API in PYQT6_API:
+    def coerce_to_qevent_type(event_value):
+        return event_value
+else:
+    def coerce_to_qevent_type(event_value):
+        return QEvent.Type(event_value)
--- a/enaml/qt/docking/layout_builder.py
+++ b/enaml/qt/docking/layout_builder.py
@@ -1,5 +1,5 @@
 #------------------------------------------------------------------------------
-# Copyright (c) 2013-2017, Nucleic Development Team.
+# Copyright (c) 2013-2022, Nucleic Development Team.
 #
 # Distributed under the terms of the Modified BSD License.
 #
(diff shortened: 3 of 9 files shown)


MatthieuDartiailh:
Great I wished I did not release 0.15.0 some hours ago. I will review ASAP. If you can try to add tests.

bburan:
@MatthieuDartiailh I discovered these only after testing by dragging the dock panes around. I see there is something called qtbot that can be used for automating tests. Pretty cool. I am not familiar with pytest-qt, so I will have to study this as I need to emulate a mouse click and drag on the dockitem titlebar. I may not be able to get to this for a few days to weeks.

MatthieuDartiailh:
Sure no rush on the tests. I would just like to improve them to be able to better trust them.

MatthieuDartiailh:
LGTM

Let me know when to merge.

enaml/qt/docking/__init__.py
```diff
@@ -1,7 +1,14 @@
...
 #
 # Distributed under the terms of the Modified BSD License.
 #
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
+from .. import QT_VERSION
+
+
+def hover_event_pos(event):
+    if QT_VERSION[0] == "6":
```
frmdstryr, line 12: Instead of checking the version in every event it'd be slightly better to define a different function

```python
if QT_VERSION[0] == "6":
    def hover_event_pos(event):
        return event.position().toPoint()
else:
    def hover_event_pos(event):
        return event.pos()

```

MatthieuDartiailh: I wanted to suggest the same thing but then confused myself about switching the qt version at runtime...

MatthieuDartiailh:
Looking for use of pos() it looks like there are several other candidate for similar fixes. I will try to make a PR based on this one if nobody beats me to it.

bburan:
@MatthieuDartiailh I don't think any of those additional `pos()` calls qualify for substitution as those are all `QMouseEvent` where `pos` is still a valid public function (https://doc-snapshots.qt.io/qt6-dev/qmouseevent.html).

bburan:
@frmdstryr Made your requested change.

MatthieuDartiailh:
Going through the Qt6 deprecation, I think we may also want to check the way we use actions and menus (in the workbench for example).

MatthieuDartiailh:
I just checked and I believe we do not use the deprecated part of Action regarding handling menus.

bburan:
@MatthieuDartiailh Hold off on merging this. I just ran across an error with PyQt6 when testing my application. Did not have a chance to look at it in more detail yet. Posting this here for reference.
```
  File "c:\users\lbhb\projects\psi2\src\enaml\enaml\qt\q_popup_view.py", line 736, in mousePressEvent
    if not path.isEmpty() and not path.contains(pos):
TypeError: arguments did not match any overloaded call:
  contains(self, QPointF): argument 1 has unexpected type 'QPoint'
  contains(self, QRectF): argument 1 has unexpected type 'QPoint'
  contains(self, QPainterPath): argument 1 has unexpected type 'QPoint'
```

bburan:
I have fixed the bug in the previous comment. However, there are additional bugs. Going through the Enaml examples and interacting with them has revealed a few more. Right now at least `drag_and_drop.enaml` (drag and drop broken), `file_dialog.enaml` (browse button broken), `flow_area.enaml` (changing flow layout), `image_view.enaml` (set to None), `ipython_console.enaml` (possibly just need to install qtconsole), `vtk_canvas.enaml` (most likely need to install vtk), `web_view.enaml`.

frmdstryr:
>  browse button broken

This should be fixed when qtpy 2.1 is released (https://github.com/spyder-ide/qtpy/pull/331).

bburan:
@MatthieuDartiailh Do you know if the `toPython` methods on the QDate and QTime objects in PyQt6 would be fixed? I found it odd that they were dropped for QDate and QTime but not QDateTime.

MatthieuDartiailh:
Thanks @bburan that will be very helpful. Regarding QDate and QTime no I have no clue. 

I am trying to finish some stuff on pegen to unblock the related PR and I will comlz back to this right after.

MatthieuDartiailh:
I started working on writing tests for the parts that require some changes under Qt6 but did not go very far due to bugs in qtest.  I will let you know when I make progress.

MatthieuDartiailh:
I made some progress with testing and made a PR against @bburan branch at https://github.com/bburan/enaml/pull/1

I will try to look at the other failures next.

MatthieuDartiailh:
For the image view I do not see a hard crash just PyQt complaining about an invalid size and I see the same on Qt5. Did you see the same behavior @bburan ?

For the flow layout it appears to be more of the float vs int issues that has been plaguing us for some time now. Testing for this will never be bulletproof I fear.

MatthieuDartiailh:
I opened #489 to track all the remaining issues. I suggest we merge this as is and I will reopen https://github.com/bburan/enaml/pull/1 as a separate PR against nucleic repo. Having small PRs will be more manageable I believe. WDYT @bburan ?

bburan:
@MatthieuDartiailh Not sure how I missed this message. Yes, I think merging as-is is fine then we can use bburan#1 as the separate PR. Agree regarding small PRs.

PR #488, opened 2022-04-07
Drop use of Qt compiled resource file

### LINKED ISSUE #487: Dock Area resource file needs to be fixed for Qt6
It looks like `enaml/qt/docking/dock_resources.py` is pretty important for getting the nice, clean hints that pop-up when attempting to dock windows (i.e., the compass rose). However, `qRegisterResourceData` appears to be missing from `PyQt6.QtCore` and I can't find any guidance on the appropriate replacement. For example, it looks like we are still supposed to use `rcc` to convert to a Python module (e.g., https://doc.qt.io/qt-6/resources.html). Yet, the file that's created imports `QtCore` from `PySide6` instead of `PyQt6`.

It seems that we still have `PySide6.QtCore.qRegisterResourceData` but not `PyQt6.QtCore.qRegisterResourceData`. If I install `PySide6`, then the compass rose image appears nicely. 

I'm a bit unsure about the best path forward here to ensure compatibility with PyQt6 as well. Thoughts?

bburan:
PyQt6 no longer supports *.rcc files. Just load the images directly.
Also, switched to QPixmap for rendering the compass rose since that
seems to be the preferred class for on-screen rendering.

Fixes #487


MatthieuDartiailh:
Nice ! I do not get why we see a segfault after pytest complete on 3.10 with pyside 6 on Linux. It is unrelated to this changes and should  not block merging this.

MatthieuDartiailh:
#486 managed to pass at least once. Hopefully it fixes the issue (I restarted the build to see if it keeps on passing).

MatthieuDartiailh:
It does not... I feel we will have to live with this one for some time except if somebody can debug locally (I personally don't have the bandwidth too). @bburan is this and #486 ready to go and do you have any other fixes pending ?

bburan:
@MatthieuDartiailh I also don't have the bandwidth to debug the pytest issue. This is likely a problem in the pytest library, but can't say for sure. I don't  have any additional Qt6 fixes pending at the moment. I think both #486 and this one are ready to go.

PR #490, opened 2022-05-02
bump qtpy minimal required version for Qt6

MatthieuDartiailh:


MatthieuDartiailh:
Ideally this needs a test before being merged

MatthieuDartiailh:
This needs a test for the dialog regression it fixes in Qt6 before merging.

MatthieuDartiailh:
I looked at testing and it looks quite a bit more complicated than I thought at first. Since the issue is deep within the compatibility offered by qtpy I will merge as is.

PR #491, opened 2022-05-02
Fix drag and drop on Qt6

MatthieuDartiailh:
This is based on #486

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
           pip install PyQt6-QScintilla
       - name: Install test requirements
         run: |
-          pip install -r test_requirements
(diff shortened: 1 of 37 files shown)


MatthieuDartiailh:
This is nearly ready to go. I will just try to add tests for the time editor and the popup view closing behavior. Testing the dock area dragging won't be possible except at a super low level and I don't think it is truly worth it.

MatthieuDartiailh:
@bburan since we have at least basic test for everything this PR covers I suggest we merge. WDYT ?

bburan:
@MatthieuDartiailh I agree. It will be difficult to test things that require dragging. For most other tests, the documentation examples  should capture many of the potential issues even if we have to test them one by one by hand. That should only be necessary for major Qt releases.

Issue #493, opened 2022-05-14
Can the expression engine provide context for errors somehow?


frmdstryr:
If an error occurs in an expression during the initialization it provides very little context as to where the error is making it difficult to debug in more complex applications.

For example this script:
```python
from enaml.core.api import Conditional
from enaml.widgets.api import Window, Container, Label


enamldef Main(Window):
    Container:
        Conditional:
            condition = None
            Label:
                text = "Hello"
```

Will generate this error:

```python-traceback
Traceback (most recent call last):
  File "/home/user/micromamba/envs/enaml/bin/enaml-run", line 33, in <module>
    sys.exit(load_entry_point('enaml', 'console_scripts', 'enaml-run')())
  File "/home/user/projects/enaml/enaml/runner.py", line 71, in main
    window.show()
  File "/home/user/projects/enaml/enaml/widgets/window.py", line 380, in show
    self.initialize()
  File "/home/user/projects/enaml/enaml/widgets/window.py", line 158, in initialize
    super(Window, self).initialize()
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 160, in initialize
    super(ToolkitObject, self).initialize()
  File "/home/user/projects/enaml/enaml/core/declarative.py", line 135, in initialize
    child.initialize()
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 160, in initialize
    super(ToolkitObject, self).initialize()
  File "/home/user/projects/enaml/enaml/core/declarative.py", line 135, in initialize
    child.initialize()
  File "/home/user/projects/enaml/enaml/core/pattern.py", line 38, in initialize
    self.refresh_items()
  File "/home/user/projects/enaml/enaml/core/conditional.py", line 74, in refresh_items
    if self.condition:
TypeError: The 'condition' member on the 'Conditional' object must be of type 'bool'. Got object of type 'NoneType' instead.
```

The trace shows the initialization stack but no context as to where the Conditional is in the tree. Since the Conditional is an instance of a node in the tree can you think of any way to provide some sort of context? 

One thought that comes to mind is having declarative do child initialization in a try except and do some sort of inspection?

MatthieuDartiailh:
Error within declarative widgets are indeed an area where we can improve much. In this specific case, having some try except in initialization may be the best solution since the error is not in the declarative part but simply in the validation of the member. However we could also try to improve error reported in the expression engine for cases when the error occurs at runtime and not at initialization time.

[...]

PR #494, opened 2022-05-17
Show declarative stack if an error occurs during initialization or change event

frmdstryr:
For #493 . It works by saving a reference to the compiler node to each declarative object and copies the filename and line number during compilation. 

Unfortunately this breaks any existing caches because of the additional parameters, I'm not sure what to do about that, it could maybe provide empty defaults for the filename and lineno?

Also it currently doesn't catch errors when manually updating an expression, eg so I'm not sure if it should be doing a try/except in the expression engine itself.

### DIFF
--- a/tests/core/test_declarative_error.py
+++ b/tests/core/test_declarative_error.py
@@ -8,6 +8,7 @@
 import sys
 
 from textwrap import dedent
+from contextlib import contextmanager
 
 import pytest
 
@@ -17,14 +18,17 @@ from enaml.widgets.api import Window
 from utils import compile_source, wait_for_window_displayed
 
 
-
+@contextmanager
 def destroy_windows():
     # Cleanup windows that do not call destroy due to an "unclean" shutdown
-    for w in list(Window.windows):
-        w.destroy()
+    try:
+        yield
+    finally:
+        for w in list(Window.windows):
+            w.destroy()
 
 
-def test_error_during_init(enaml_qtbot):
+def test_error_pattern_node_included_init(enaml_qtbot):
     source = dedent("""\
     from enaml.core.api import Conditional
     from enaml.widgets.api import Window, Container, Label
@@ -38,11 +42,33 @@ def test_error_during_init(enaml_qtbot):
 
     """)
     tester = compile_source(source, 'Main')()
-    with pytest.raises(DeclarativeError) as excinfo:
-        tester.show()
+    with destroy_windows():
+        with pytest.raises(DeclarativeError) as excinfo:
+            tester.show()
     assert 'line 4, in Main' in excinfo.exconly()
     assert 'line 5, in Container' in excinfo.exconly()
     assert 'line 6, in Conditional' in excinfo.exconly()
+
+
+def test_error_multiple_same_cls_init(enaml_qtbot):
+    source = dedent("""\
+    from enaml.widgets.api import Window, Container, Label
+
+    enamldef Main(Window): main:
+        Container:
+            Label: lbl1:
+                text = False
+            Label: lbl2:
+                text = "Ok"
+    """)
+    tester = compile_source(source, 'Main')()
+    with destroy_windows():
+        with pytest.raises(DeclarativeError) as excinfo:
+            tester.show()
+    assert 'line 3, in Main' in excinfo.exconly()
+    assert 'line 4, in Container' in excinfo.exconly()
+    # Make sure it is not showing the error for Label 2
+    assert 'line 5, in Label' in excinfo.exconly()
 
 
 def test_error_during_read_expr(qt_app, qtbot):
@@ -55,8 +81,9 @@ def test_error_during_read_expr(qt_app, qtbot):
                 text = undefined_variable
     """)
     tester = compile_source(source, 'Main')()
-    with pytest.raises(DeclarativeError) as excinfo:
-        tester.show()
+    with destroy_windows():
+        with pytest.raises(DeclarativeError) as excinfo:
+            tester.show()
 
     assert 'line 3, in Main' in excinfo.exconly()
     assert 'line 4, in Container' in excinfo.exconly()
@@ -91,11 +118,9 @@ def test_error_during_event(qt_app, qtbot):
     tester.show()
     wait_for_window_displayed(qtbot, tester)
 
-    try:
+    with destroy_windows():
         with pytest.raises(DeclarativeError) as excinfo:
             tester.widget.button.clicked(True)
-    finally:
-        destroy_windows()
 
     assert 'line 13, in Main' in excinfo.exconly()
     assert 'line 17, in MyWidget' in excinfo.exconly()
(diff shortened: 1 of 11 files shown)


MatthieuDartiailh:
This sounds great. Could you post the resulting traceback for reference ?

I will do my best to review this week.

frmdstryr:
The error when clicking the button for this

```python
from enaml.core.api import Conditional


enamldef MyContainer(Container): inside_my_container:
    Label: my_label:
        text = "Hello"
    Conditional: cond:
        condition = True#None
        PushButton: btn:
            text = "Blow up"
            clicked :: cond.condition = None

enamldef Main(Window): main:
    Container: container:
        Label: label1:
            text = "Test"
        MyContainer: my_container:
            pass

```

Will look like this:

```python-traceback
  File "/home/user/projects/enaml/enaml/core/declarative_meta.py", line 75, in declarative_change_handler
    engine.write(owner, change['name'], change)
  File "/home/user/projects/enaml/enaml/core/expression_engine.py", line 210, in write
    pair.writer(owner, name, change)
  File "/home/user/projects/enaml/enaml/core/standard_handlers.py", line 82, in __call__
    call_func(func, (), {}, scope)
  File "tests/example.enaml", line 1, in f
    from enaml.core.api import Conditional
TypeError: The 'condition' member on the 'Conditional' object must be of type 'bool'. Got object of type 'NoneType' instead.

The above exception was the direct cause of the following exception:

  File "/home/user/projects/enaml/enaml/qt/qt_abstract_button.py", line 83, in on_clicked
    self.declaration.clicked(checked)
  File "/home/user/projects/enaml/enaml/core/declarative_meta.py", line 79, in declarative_change_handler
    raise DeclarativeError(owner, e) from e
enaml.core.declarative_meta.DeclarativeError: 
  File "tests/example.enaml", line 14, in Main
  File "tests/example.enaml", line 15, in Container
  File "tests/example.enaml", line 18, in MyContainer
  File "tests/example.enaml", line 10, in PushButton
Aborted
```

It'd be nice if it could show the source lines all the way up and actual expression but I this at least shows which block.

tests/core/test_declarative_error.py
```diff
@@ -0,0 +1,110 @@
...
+    """)
+    tester = compile_source(source, 'MyWindow')()
+    with pytest.raises(DeclarativeError) as excinfo:
+        tester.show()
+    assert 'line 4, in MyWindow' in excinfo.exconly()
+    assert 'line 5, in Container' in excinfo.exconly()
+    assert 'line 6, in Conditional' in excinfo.exconly()
+
+
+@pytest.mark.qt_no_exception_capture
```
MatthieuDartiailh, line 47: This feels excessive. We are not testing Qt here so simply manually firing the event would be enough.

frmdstryr: Indeed.

MatthieuDartiailh:
@frmdstryr I am a bit swamped at the moment so please ping me when you want a review.

frmdstryr:
Unfortunately I don't really like how the ` DeclarativeError() from e` works... Most of the time I have to scroll up to the original error. Any ideas for an alternate approach?

frmdstryr:
For reasons I don't quite understand yet an error in a default value is not propagated up with this branch. Eg this is simply ignored.

```python
from enaml.widgets.api import Window, Container, Label

enamldef MyWindow(Window):
    Container:
        Label:
            text = undefined_variable
```

MatthieuDartiailh:
That does not make any sense. What happens if you simply reraise the same exception rather than the declarative one.

frmdstryr:
If I raise a ValueError then it works so maybe it's something with validation.

frmdstryr:
Oh sorry, it's because it is doing 'pass' when DeclarativeError is raised... https://github.com/nucleic/enaml/pull/494/files#diff-68b4f23d01f700d0980fdb56a264c7bb6313fcd987739a0e7e27e9bf467d1882R189

MatthieuDartiailh:
I do lot have an alternate idea. I am not surprised you need to know what failed to fix the issue bit knowing where to look is really helpful.

Currently you do not use the compiler node you store and because you the the widget parent you can miss element which are not widgets (Include, Looper,...). I believe you could do better by checking if the node of the parent does contain the node of the child and if not look in the children of the parent for an element whose node does contain the child node.

I would also be interested in seeing some tests for templates and template instances.

MatthieuDartiailh:
Tests for templates are still missing but is this otherwise ready for review ?

frmdstryr:
It still doesn't work right for templates but feedback is welcome.

frmdstryr:
This template 

```python
from enaml.widgets.api import Window, Container, Html

template Panel(Content):
    Container:
        Content:
            pass

enamldef HtmlContent(Html):
    source = False

enamldef Main(Window): main:
    Panel(HtmlContent):
        pass
```

Currently gives this  

```
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 224, in activate_proxy
    child.activate_proxy()
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 220, in activate_proxy
    self.activate_top_down()
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 241, in activate_top_down
    self.proxy.activate_top_down()
  File "/home/user/projects/enaml/enaml/qt/qt_toolkit_object.py", line 74, in activate_top_down
    self.init_widget()
  File "/home/user/projects/enaml/enaml/qt/qt_html.py", line 40, in init_widget
    self.set_source(self.declaration.source)
TypeError: The 'source' member on the 'HtmlContent' object must be of type 'str'. Got object of type 'bool' instead.

The above exception was the direct cause of the following exception:

  File "/home/user/micromamba/envs/enaml/bin/enaml-run", line 33, in <module>
    sys.exit(load_entry_point('enaml', 'console_scripts', 'enaml-run')())
  File "/home/user/projects/enaml/enaml/runner.py", line 71, in main
    window.show()
  File "/home/user/projects/enaml/enaml/widgets/window.py", line 382, in show
    self.activate_proxy()
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 224, in activate_proxy
    child.activate_proxy()
  File "/home/user/projects/enaml/enaml/widgets/toolkit_object.py", line 228, in activate_proxy
    raise DeclarativeError(child, e) from e
enaml.core.declarative_meta.DeclarativeError: 
  File "tests/example.enaml", line 11, in Main
    enamldef Main(Window): main:
  File "tests/example.enaml", line 12, in template
    Panel(HtmlContent):
  File "tests/example.enaml", line 4, in Container
    Container:
  File "tests/example.enaml", line 5, in HtmlContent
    Content:
TypeError: The 'source' member on the 'HtmlContent' object must be of type 'str'. Got object of type 'bool' instead.
```

I'm not sure how to get the template class name from line 12. Also it's inserting the HtmlContent node into the template as Content so the trace needs to somehow capture that...

MatthieuDartiailh:
Looking a bit at the compiler it looks like currently the information we would need is discarded. I will try to dig deeper later. Also I am wondering if we could make it more explicit when we move across different definitions.

frmdstryr:
It appears that linecache becomes stale if a file is reloaded. Not sure if that needs addressed?

MatthieuDartiailh:
I have not forgotten about this but I am swamped at the moment (with bytecode for 3.11 and hence enaml for 3.11 at the moment).

frmdstryr:
I don't like this approach.

MatthieuDartiailh:
I can understand your opinion. Still I wish we could do something to improve tracebacks.

PR #496, opened 2022-06-13
core: use atom.add_member to add members after a class is defined

MatthieuDartiailh:
This will require atom >= 0.9


frmdstryr:
I have been using an old version for a while and just updated a few days ago (to atom's git master) and noticed that dock themes no longer load at startup.  I can reproduce it by running the dock area example.  

image

If I downgrade to atom 0.8.1 and revert this commit it works as expected...

image

Are you able to reproduce it?

MatthieuDartiailh:
Yes I can reproduce. Not sure when I will have time to investigate though.

PR #503, opened 2022-08-12
qt: fix loading dock images

MatthieuDartiailh:
We test directly the loading since we cannot test D&D on the dock area

This is extremely annoying and I will do my best to release a patch version ASAP

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,20 @@
+[run]
+branch = True
+include =
+    */enaml/*
+omit =
+    */enaml/qt/docking/dock_images/*
+    */tests/*
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain if tests don't hit defensive assertion code:
+    raise NotImplementedError
+    pass
+
+    # Dont't complain about type checking
+    if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
ping @bburan since you worked on the previous patch

frmdstryr:
I think the test is just accessing the member not triggering the factory?  It would be good to call the layout method to get a little coverage.

MatthieuDartiailh:
I meant to access the guide on an instance... I added calls to layout

bburan:
What is the bug here? The only fix I see is to get rid of the import for QDir?

MatthieuDartiailh:
It has to do with the fact that we need a valid package for `path`. And without the extra `__init__` things were going really wrong.

bburan:
@MatthieuDartiailh I see. For some odd reason I can't replicate the issue. On Python 3.9 at least it seems to automatically add a `__init__.py` file when installing to `site-packages`. Perhaps this bug is dependent on the version of setuptools? I've only just started learning about the new approach to Python packaging (i.e., putting everything in pyproject.toml).

Either case, I see no reason to hold off on this PR.

bburan:
Just tested on a clean setup of Python 3.8 and I can replicate the bug. Applying this patch fixes the bug. Interesting. Probably a newer version of setuptools or setuptools_scm (that gets installed on Python 3.9) is smart enough to auto-create the `__init__.py` file? Either case, I can definitely confirm this patch fixes it.

MatthieuDartiailh:
Thanks for testing thoroughly. I will merge and try to do a patch release ASAP.

PR #504, opened 2022-08-18
Typo

davirenno:
At line 44, it is written "protoype". I suggested the change to "prototype", if it is the intention.


MatthieuDartiailh:
Thanks

PR #505, opened 2022-08-19
Prepare release 0.15.2

MatthieuDartiailh:
I reverted d47be167 that requires the development branch of atom which is not ready for a new release.


MatthieuDartiailh:
CIs failure are expected because we test against atom main branch. The wheel building process so we are fine.

Issue #510, opened 2022-10-20
[DOC] How to use async/await function in Enaml ?


Kochise:
I'd like to use `async`/`await` functions and avoid blocking the UI.

Is there a "progress bar" available for pending operations ?

MatthieuDartiailh:
There is a progress bar widget but I am not sure this what you are looking for. Using `qasync` you may be able to get what you want.

Kochise:
`qasync` ?

MatthieuDartiailh:
If you want to use `async / await` you need an asyncio compatible eventloop. `qasync` implement one on top of the Qt event loop.

Kochise:
So not directly inside a `enaml` object ?

MatthieuDartiailh:
You can write async func in an Enaml object. At one point you will need to start an asyncio.Task and this requires an asyncio event loop. @frmdstryr may be able to provide a more in depth description since he has been doing more work using async.

Kochise:
Ok thank, looking forward to it.

Kochise:
Works with `qasync`, thanks.

Closing...

PR #511, opened 2022-10-20
Add async example

frmdstryr:
For #510 . A similar approach can be used with other qt / asyncio libraries.

### DIFF
--- a/examples/tutorial/async/async.py
+++ b/examples/tutorial/async/async.py
@@ -1,0 +1,85 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2022, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ------------------------------------------------------------------------------
+import asyncio
+import os
+import sys
+from inspect import iscoroutine
+from queue import Empty, Queue
+
+import enaml
+from atom.api import Bool, Typed
+from enaml.qt.qt_application import QtApplication
+
+try:
+    from asyncqtpy import QEventLoopPolicy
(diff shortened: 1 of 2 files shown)


examples/tutorial/async/async.py
```diff
@@ -0,0 +1,85 @@
...
+import sys
+from inspect import iscoroutine
+from queue import Empty, Queue
+
+import enaml
+from atom.api import Bool, Typed
+from enaml.qt.qt_application import QtApplication
+
+try:
+    from asyncqtpy import QEventLoopPolicy
```
MatthieuDartiailh, line 19: astncqtpy readme explicitly suggests to use qasync in its README, so would it not be better to use it instead.

frmdstryr: asyncqtpy is a fork I made that uses qtpy.  From what I recall qasync did not work.

MatthieuDartiailh: I will check what I did but I have used qasync with enaml but I may have taken some shortcuts

Kochise:
https://github.com/CabbageDevelopment/qasync

examples/tutorial/async/async.py
MatthieuDartiailh: I could get this to work trivially with qasync (it probably needs to be imported after qtpy). So I allowed the example to work with it. Let me know if this is ok with you and I will merge after adding an entry in the release notes.

frmdstryr: Fine with me

MatthieuDartiailh:
Could you rebase on main since to get green CIs ?

Kochise:
Works with `qasync`, thanks.

PR #512, opened 2022-10-23
button_group

mervecelik11:


MatthieuDartiailh:
Could you motivate your changes ?

MatthieuDartiailh:
Do not worry about the CI failure I will take care of it.

mervecelik11:
Thank you

MatthieuDartiailh:
Reading through your changes it seems you have misinterpreted the goal of the example, which means it can be improved.

This example demonstrate how radio buttons living in different containers can still be grouped. Radio buttons living in the same container are grouped by default which is not demonstrated by this example and maybe it should. Your changes make use of this mechanism but completely remove the original intent of the example. 

We should try to expand the example to do both: demonstrate the default behavior and the more flexible one.

PR #514, opened 2022-12-19
Add python 3.11 support

frmdstryr:
This updates the code generator/compiler to work with Python 3.11, it builds on the pegen branch because the main branch just fails due to invalid col offsets in the ast. 

- It seems like the python docs are wrong with MAKE_FUNCTION as there is now no need to push a qualname (https://github.com/python/cpython/issues/93270).
- Since POP_BLOCK was removed, I'm not sure if `try_squash_raise` can still be implemented?
- There is still a weird problem with decl funcs with default args. Inspect shows the signature is correct and funchelper shows the correct number of defaults but for some reason it does not use the defaults?

On a positive note I get about a 10-15% speedup with this on enaml-web :).

### DIFF
--- a/enaml/core/code_generator.py
+++ b/enaml/core/code_generator.py
@@ -323,14 +323,14 @@ class CodeGenerator(Atom):
             bc.Instr("LOAD_BUILD_CLASS"),               # TOS -> builtins.__build_class__
         )
 
-    def make_function(self, n_defaults=0, name=None):
+    def make_function(self, flags=0, name=None):
         """ Make a function from a code object on the TOS.
 
         """
         if not PY311:
             self.load_const(name)
         self.code_ops.append(                           # TOS -> qual_name -> code -> defaults
-            bc.Instr("MAKE_FUNCTION", n_defaults),      # TOS -> func
+            bc.Instr("MAKE_FUNCTION", flags),           # TOS -> func
         )
 
     def push_null(self):
--- a/enaml/core/code_generator.py
+++ b/enaml/core/code_generator.py
@@ -1,5 +1,5 @@
 #------------------------------------------------------------------------------
-# Copyright (c) 2013-2021, Nucleic Development Team.
+# Copyright (c) 2013-2023, Nucleic Development Team.
 #
 # Distributed under the terms of the Modified BSD License.
 #
--- a/enaml/core/code_generator.py
+++ b/enaml/core/code_generator.py
@@ -362,7 +362,7 @@ class CodeGenerator(Atom):
                 # kwargs_name should be a tuple listing the keyword
                 # arguments names
                 # TOS -> func -> args -> kwargs -> kwargs_names
-                op, arg = "CALL_FUNCTION_KW", n_args+n_kwds
+                op, arg = "CALL_FUNCTION_KW", n_args + n_kwds
             else:
                 op, arg = "CALL_FUNCTION", n_args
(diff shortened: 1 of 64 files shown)


enaml/core/code_generator.py
```diff
@@ -383,6 +430,9 @@ def try_squash_raise(self):
         the code, rather than any function called by the code.
 
         """
+        if PY311:
+            yield
+            return  # TODO: Is this still possible?
```
MatthieuDartiailh, line 435: Using TryBegin and TryEnd pseudo instructions this should be feasible. I can give it a try but I may prioritize getting the pegen branch in first since we need it to.

enaml/core/code_generator.py
```diff
@@ -308,27 +323,49 @@ def load_build_class(self):
...
 
-    def make_function(self, n_defaults=0):
+    def make_function(self, n_defaults=0, name=None):
         """ Make a function from a code object on the TOS.
 
         """
+        if not PY311:
+            self.load_const(name)
         self.code_ops.append(                           # TOS -> qual_name -> code -> defaults
             bc.Instr("MAKE_FUNCTION", n_defaults),      # TOS -> func
```
MatthieuDartiailh, line 333: Under 3.11 MAKE_FUNCTION does not take the number of default anymore but a code (see https://docs.python.org/3.11/library/dis.html#opcode-MAKE_FUNCTION). This probably explain your issue.

frmdstryr: It looks like this has been the case ever since 2.7, anyways there is only one call to make_function with defaults and it happens to be `0x01` so that is why it worked.

MatthieuDartiailh: I thought it could be the origin of your default argument issue. I will have to look elsewhere (possibly in the C helpers)

MatthieuDartiailh:
Could you rebase this on main now that #474 is in ? (Cherry picking your changes might be easier) Also I apologize for the last two commits, I forgot to switch branch.

MatthieuDartiailh:
So I have an implementation for try_squash_raise. I just need to audit the usages to ensure we never have a try/except inside the try_squash_raise.

I will look at the kwarg default next.

MatthieuDartiailh:
Thanks for rebasing. I will push my solution for try_squah_raise later today. The issue with function default lies in CPython itself. PyEval_EvalCodeEx started using _PyFunction_FromConstructor in 3.11 but this function does not pass the defaults found in the frame constructor... I will open an issue on CPython repo.

MatthieuDartiailh:
CPython issue reported at https://github.com/python/cpython/issues/101072

MatthieuDartiailh:
Still blocked on the CPython issue, the PR I made is pending review.

MatthieuDartiailh:
The cpython issue has been closed and the fix will be in 3.11.2 ! I will try to have everything ready by then.

MatthieuDartiailh:
3.11.2 is available on GHA (I misremembered the release date), let's see if it works !

frmdstryr:
Looks like it still is using 3.11.1

MatthieuDartiailh:
Yes and I find it extremely annoying ! We will have to wait a tiny bit longer it seems.

MatthieuDartiailh:
Thanks for your work @frmdstryr

frmdstryr:
Thank you!

PR #517, opened 2023-01-26
Implement hash and eq for bound decl func

frmdstryr:
Currently if you use observe with a decl func it is impossible to remove the observer unless you clear all observers. This fixes that. 

A simple example to show the problem.  

```python
from atom.api import Atom, Str
from enaml.widgets.api import Window, Container, Label, Field, PushButton


class Model(Atom):
    output = Str()


    attr model = Model()
    func decl_print(x):
        print(x)

    Container:
        PushButton:
            text = "Add observer"
            clicked :: model.observe("output", decl_print)
        PushButton:
            text = "Remove observer"
            clicked :: model.unobserve("output", decl_print)
        Field:
            submit_triggers = ["auto_sync"]
            text := model.output

```

The hash function is based on cpythons method_hash (see https://github.com/python/cpython/blob/81f7359f67a7166d57a10a3d5366406d9c85f1de/Objects/classobject.c#L298)

### DIFF
--- a/enaml/src/declarative_function.cpp
+++ b/enaml/src/declarative_function.cpp
@@ -360,6 +360,48 @@ BoundDMethod__call__( BoundDMethod* self, PyObject* args, PyObject* kwargs )
 
 
 PyObject*
+BoundDMethod_richcompare( BoundDMethod* self, PyObject* other, int opid  )
+{
+    if( opid == Py_EQ )
+    {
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
Can you add a changelog entry ?

MatthieuDartiailh:
Thanks

Issue #521, opened 2023-04-14
TypeError when using try/except with "as" in Enaml code


bburan:
The following code block in any Enaml file:
```
try:
    pass
except Exception as e:
    pass
```
Results in the following traceback:
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Scripts\enaml-run.exe\__main__.py", line 7, in <module>
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\runner.py", line 44, in main
    code = EnamlCompiler.compile(ast, enaml_file_path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_compiler.py", line 179, in compile
    return compiler.visit(node)
           ^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_ast.py", line 418, in visit
    result = visitor(node, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_compiler.py", line 195, in visit_Module
    self.visit(item)
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_ast.py", line 418, in visit
    result = visitor(node, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\enaml_compiler.py", line 213, in visit_PythonModule
    cg.insert_python_block(node.ast)
  File "c:\Users\lbhb\anaconda3\envs\psi-nafc\Lib\site-packages\enaml\core\code_generator.py", line 573, in insert_python_block
    bc_code = cfg.to_bytecode()
              ^^^^^^^^^^^^^^^^^
[...]

PR #528, opened 2023-05-23
Fix code generation for inserted Python block on Python 3.10+

MatthieuDartiailh:
On Python 3.11 in particular one can often find a last block which acts as an error handling block which should not be connected to the code we want to append after the block.


MatthieuDartiailh:
Rely on https://github.com/MatthieuDartiailh/bytecode/pull/120 against which we are testing to ensure the fix there is correct.

MatthieuDartiailh:
closes #521

PR #546, opened 2024-03-23
Add initial enamldef const storage

frmdstryr:
This makes the parser allow a "const" in an enamldef or child def that uses atom's `ReadOnly` member so it cannot be changed after the initial value is set. The only other difference from an attr is that the parser requires a default expression.

The existing parser (pegen not the ply one) actually allows "const" in an enamldef but not in a child def but throws an error because the visitor is not implemented in the block compiler.

I copied a lot of code in the gen_const_expr, idk if that can/should be cleaned up. It's mainly an inlined gen_storage_expr and gen_operator_expr.

An example use case...
```python

    const locked: bool = True
    Container:
        Label:
            text = "Locked" if locked else "Unlocked"
        PushButton:
            text = "Try to unlock"
            clicked ::
                try:
                    window.locked = False
                except TypeError:
                    print("Nope!")

```

Any feedback is welcome.

### DIFF
--- a/enaml/core/block_compiler.py
+++ b/enaml/core/block_compiler.py
@@ -111,6 +111,10 @@ class FirstPassBlockCompiler(BaseBlockCompiler):
         # Grab the index of the parent node for later use.
         self.aux_index_map[node] = self.parent_index()
 
+    def visit_ConstExpr(self, node):
+        # Grab the index of the parent node for later use.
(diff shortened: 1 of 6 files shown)


sccolbert:
Does this really gain us anything in practice?

I am generally against adding to the syntax unless it introduces an
entirely new concept to the language.

PR #552, opened 2024-05-22
Add nonblocking open support to Dialog

yajaru:
Add a non-blocking 'open' to Dialog so that applications may continue to run their existing event loops while the dialog is open.
Applications that wish to receive results from the Dialog can do so via the existing 'finished' event.

### DIFF
--- a/enaml/widgets/dialog.py
+++ b/enaml/widgets/dialog.py
@@ -22,6 +22,9 @@ class ProxyDialog(ProxyWindow):
     def exec_(self):
         raise NotImplementedError
 
+    def open(self):
+        raise NotImplementedError
+
     def done(self, result):
         raise NotImplementedError
 
@@ -67,6 +70,16 @@ class Dialog(Window):
         if not self.proxy_is_active:
             self.activate_proxy()
         return self.proxy.exec_()
+
+    def open(self):
+        """ Launch the dialog as a modal window.
+
+        """
+        if not self.is_initialized:
+            self.initialize()
+        if not self.proxy_is_active:
+            self.activate_proxy()
+        return self.proxy.open()
 
     def done(self, result):
         """ Close the dialog and set the result value.
--- a/enaml/qt/qt_dialog.py
+++ b/enaml/qt/qt_dialog.py
@@ -102,6 +102,12 @@ class QtDialog(QtWindow, ProxyDialog):
         """
         return bool(self.widget.exec_())
 
+    def open(self):
+        """ Launch the dialog as a modal window.
+
+        """
+        self.widget.open()
+
     def done(self, result):
         """ Close the dialog and set the result code.
 


MatthieuDartiailh:
@sccolbert I am interested in the historical reasons you had not to provide such a mode initially.

yajaru:
> @sccolbert I am interested in the historical reasons you had not to provide such a mode initially.

My guess is that open wasn't available when enaml on QT was being made but, this is pure speculation.
image

sccolbert:
I'm pretty sure you can call `.show()` on a dialog and it will open and
return immediately. However, it won't be modal.

yajaru:
> I'm pretty sure you can call `.show()` on a dialog and it will open and return immediately. However, it won't be modal.

`show` does exist on Dialog by virtue of it existing for all enaml widgets (which I am leery on, but that is a different conversation). However, losing the modal behavior is a significant loss given that this is a dialog and `exec_` has the modal behavior we are trying to separate from the blocking.

Mostly I'm just exposing the api that QT recommends instead of `exec` (see [https://doc.qt.io/qt-6/qdialog.html#exec](https://doc.qt.io/qt-6/qdialog.html#exec), specifically the note) as it has the closest semantics to `exec`.

sccolbert:
I'm not arguing against the addition, just pointing out an option that
already exists in case it worked for you.

Historically, the Dialog class was written when we also supported the wx
backend, so that Qt method may have been overlooked for compatibility
reasons.

yajaru:
For what it's worth, I took a quick perusal of some other toolkits to see how they handle dialogs.

[GTK3](https://docs.gtk.org/gtk3/class.Dialog.html) has a similar blocking/non-blocking split to QT (with the default falling on the non-blocking side)
[GTK4](https://docs.gtk.org/gtk4/class.Dialog.html) drops the blocking API and only leaves the non-blocking version.

[Win32](https://learn.microsoft.com/en-us/windows/win32/dlgbox/using-dialog-boxes) supports both blocking and non-blocking.
[WinUI 3](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.contentdialog?view=windows-app-sdk-1.5) only supports non-blocking

[AppKit](https://developer.apple.com/documentation/appkit/nsalert) supports both blocking and non-blocking
[SwiftUI](https://developer.apple.com/documentation/swiftui/modal-presentations) only supports non-blocking

Note that I'm not a developer in any of these so, it is very possible that I have gotten something wrong or misunderstood the docs. That being said, the world looks to be moving towards non-blocking dialogs exclusively for newer toolkits, which lends credence to this change or something in the same vein.

MatthieuDartiailh:
Thanks for the detailed explanation. Given it means aligning with most toolkit, this addition looks fine to me. 

Could you add some tests and an entry to the changelog ?

yajaru:
I've written a test but no clue if it works though. Unfortunately, I have not been able to get enaml building on my machine due to dependency issues so, I'm unable to run the test to check if it works.
If there are any dumb errors, then squash will have to do for it.

yajaru:
Ok, this should be ready to go. I would personally squash this down b/c there are intermediate commits due to debugging difficulties.

Some notes:
- I only wrote a test over the new async Dialog path. This is b/c I'm not sure how best to test the sync Dialog given that it blocks on the show. This should be fine as the sync path is untouched by my changes.
- There should also changes made to the stdlib dialogs to expose the async interface for them as well. I'll leave that as an exercise for later.
- I encountered some weirdness that I'm not sure if I can class as a bug or not. Basically, when we close the Dialog, it sticks around in the windows set and does not remove itself. I'm not sure if this is intended but, once we not longer have a live window for the dialog and all the references to it are gone, I would expect it to remove itself from the set.

sccolbert:
Do other windows automatically remove themselves from the set, or is it
something in their "destroy" method (or another signal handler), that
removes them.

IIRC, that list is not keeping weakrefs.

yajaru:
The addition and removal to and from the windows set is done in the initialize and destroy methods. So I guess the issue is that we are not fully executing the lifecycle of the Dialog with the way I have done it here (which is creating a dialog object outside of the tree).
https://github.com/nucleic/enaml/blob/ca6cf4a6d80eab455cbe8201f9251b46bd185fa7/enaml/widgets/window.py#L152C5-L168C37

This raises the question of what is the proper way to create and show a Dialog given that it is often done in a callback from some user initiated action like a button.

enaml/widgets/dialog.py
```diff
@@ -68,6 +71,16 @@ def exec_(self):
             self.activate_proxy()
         return self.proxy.exec_()
 
+    def open(self):
+        """ Launch the dialog as a modal window.
```
MatthieuDartiailh, line 75: You could specify that this call is non blocking to make the difference with exec_

MatthieuDartiailh (commented):
Since you did not touch the closing logic I fail to see why the dialog is not removed from the set. What may happen is that Qt does not process all the required events for the closing to complete. They are some testing utilities to process events till a window is closed, could you try using them ?

enaml/widgets/dialog.py
yajaru: Done.

yajaru:
That was the ticket it seems; I needed wait the QT loop for a bit since the callback to do the destroy was deferred on it.

MatthieuDartiailh:
Tests failures are only Codecov failing to retrieve the upload_token. So all good. Could you add an entry into the changelog and I will merge.

Issue #561, opened 2025-01-03
Degrated performance with high number of subscription observers


frmdstryr:
I have an enaml-web application that renders an interactive table. There is an underlying atom data model that allows multiple users to view the same data in real time and see live updates (some users can modify/move items in the table and all other users see those changes). 

Unfortunately as the number of users grows (to even a small number like 10) the app becomes very laggy with inputs taking several seconds (much longer when cprofile is running). The table displays about 200 items in a 20 x 40 table which does not seem too excessive. Since each user viewing the page sees their own copy so there are quite a few observers watching the same objects.

I used cProfile to capture a trace when one user moves an item from one cell to another which triggers a lot of changes. 

```python-traceback
    29394177 function calls (29381118 primitive calls) in 72.940 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        3    0.263    0.088   71.053   23.684 /home/user/app/views/table.enaml:347(move_cell)
 1139/332    0.005    0.000   70.487    0.212 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/standard_tracer.py:44(__call__)
 1139/332    0.014    0.000   70.485    0.212 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/expression_engine.py:214(update)
1187/1179    0.008    0.000   70.129    0.059 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/standard_handlers.py:91(__call__)
1591/1212    0.006    0.000   70.113    0.058 {built-in method enaml.core.funchelper.call_func}
     1505    0.004    0.000   68.399    0.045 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/standard_tracer.py:159(return_value)
     1505    0.139    0.000   68.395    0.045 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/standard_tracer.py:100(finalize)
    89770   34.559    0.000   68.247    0.001 {method 'observe' of 'atom.catom.CAtom' objects}
      318    0.001    0.000   35.711    0.112 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/compiler_helpers.py:745(wrapper)
 28113376   34.062    0.000   34.062    0.000 /home/user/micromamba/envs/app/lib/python3.10/site-packages/enaml/core/standard_tracer.py:35(__bool__)
[...]

PR #562, opened 2025-01-06
Move standard_tracer's subscription observer to cpp

frmdstryr:
Moves the subscription observer in standard_tracer.py to cpp. 

This gives a considerable performance increase if there are a lot of changes going on (~60%) but for most code it will likely only be a small improvement. My application sped up about 30% (even after fixing the original problem).

Ref https://github.com/nucleic/enaml/issues/561

### DIFF
--- a/enaml/src/subscription_observer.cpp
+++ b/enaml/src/subscription_observer.cpp
@@ -125,13 +125,13 @@ SubscriptionObserver_call( SubscriptionObserver* self, PyObject* args, PyObject*
         cppy::ptr engine( owner.getattr("_d_engine") );
         if ( !engine.is_none() )
         {
-            cppy::ptr args( PyTuple_New( 2 ) );
-            if( !args )
+            cppy::ptr update_args( PyTuple_New( 2 ) );
+            if( !update_args )
                 return 0;
-            PyTuple_SET_ITEM( args.get(), 0, cppy::incref( owner.get() ) );
-            PyTuple_SET_ITEM( args.get(), 1, cppy::incref( self->name ) );
+            PyTuple_SET_ITEM( update_args.get(), 0, cppy::incref( owner.get() ) );
+            PyTuple_SET_ITEM( update_args.get(), 1, cppy::incref( self->name ) );
             cppy::ptr update( engine.getattr( "update" ) );
-            return update.call( args );
+            return update.call( update_args );
         }
     }
     Py_RETURN_NONE;
--- a/enaml/src/subscription_observer.cpp
+++ b/enaml/src/subscription_observer.cpp
@@ -129,9 +129,15 @@ SubscriptionObserver_call( SubscriptionObserver* self, PyObject* args, PyObject*
             return 0;
         if ( !engine.is_none() )
         {
-            PyObject* call_args[] = { engine.get(), owner.get(), self->name };
-            size_t nargsf = 2 | PY_VECTORCALL_ARGUMENTS_OFFSET;
-            return PyObject_VectorcallMethod(update_str.get(), call_args, nargsf, 0);
+            cppy::ptr update_args( PyTuple_New( 2 ) );
+            if( !update_args )
+                return 0;
+            PyTuple_SET_ITEM( update_args.get(), 0, cppy::incref( owner.get() ) );
+            PyTuple_SET_ITEM( update_args.get(), 1, cppy::incref( self->name ) );
+            cppy::ptr update( engine.getattr( update_str.get() ) );
+            if ( !update )
+                return 0;
+            return update.call( update_args );
         }
     }
     Py_RETURN_NONE;
--- a/enaml/src/subscription_observer.cpp
+++ b/enaml/src/subscription_observer.cpp
@@ -142,28 +142,6 @@ SubscriptionObserver_call( SubscriptionObserver* self, PyObject* args, PyObject*
 }
 
 
-PyObject*
-SubscriptionObserver_richcompare( SubscriptionObserver* self, PyObject* other, int opid )
-{
-    if( opid == Py_EQ )
-    {
-        if( SubscriptionObserver::TypeCheck( other ) )
-        {
-            SubscriptionObserver* so_other = reinterpret_cast<SubscriptionObserver*>( other );
-            if(
-                PyObject_RichCompareBool( self->atomref, so_other->atomref, Py_EQ )
-                && PyObject_RichCompareBool( self->name, so_other->name, Py_EQ )
-            )
-            {
-                Py_RETURN_TRUE;
-            }
-        }
-        Py_RETURN_FALSE;
-    }
-    Py_RETURN_NOTIMPLEMENTED;
-}
-
-
 PyDoc_STRVAR(SubscriptionObserver__doc__,
     "SubscriptionObserver(owner, name)\n\n"
     "An observer object which manages a tracer subscription.\n"
@@ -215,7 +193,6 @@ static PyType_Slot SubscriptionObserver_Type_slots[] = {
     { Py_tp_clear, void_cast( SubscriptionObserver_clear ) },              /* tp_clear */
     { Py_tp_call, void_cast( SubscriptionObserver_call ) },                /* tp_call */
     { Py_tp_doc, cast_py_tp_doc( SubscriptionObserver__doc__ ) },          /* tp_doc */
-    { Py_tp_richcompare, void_cast( SubscriptionObserver_richcompare ) },  /* tp_richcompare */
     { Py_nb_bool, void_cast( SubscriptionObserver__bool__ ) },             /* nb_bool */
     { Py_tp_getset, void_cast( SubscriptionObserver_getset ) },            /* tp_getset */
     { Py_tp_new, void_cast( SubscriptionObserver_new ) },                  /* tp_new */
--- a/enaml/src/subscription_observer.cpp
+++ b/enaml/src/subscription_observer.cpp
@@ -23,6 +23,8 @@ namespace enaml
 
 // ptr to atom.api.atomref
 static cppy::ptr atomref;
+static cppy::ptr d_engine_str;
+static cppy::ptr update_str;
 
 // POD struct - all member fields are considered private
 struct SubscriptionObserver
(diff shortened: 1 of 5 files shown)


enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+int
+SubscriptionObserver_traverse( SubscriptionObserver* self, visitproc visit, void* arg )
+{
+    Py_VISIT( self->atomref );
+    Py_VISIT( self->name );
+    #if PY_VERSION_HEX >= 0x03090000
+    // This was not needed before Python 3.9 (Python issue 35810 and 40217)
+    Py_VISIT(Py_TYPE(self));
+    #endif
```
MatthieuDartiailh, line 102: I dropped support for 3.8 and 3.9 so this can be simplified.
```suggestion
    Py_VISIT(Py_TYPE(self));
```

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,335 @@
...
+    SubscriptionObserver_clear( self );
+    Py_TYPE(self)->tp_free( reinterpret_cast<PyObject*>( self ) );
+}
+
+
+int
+SubscriptionObserver__bool__( SubscriptionObserver* self )
+{
+    cppy::ptr atomrefptr( cppy::incref( self->atomref ) );
+    return atomrefptr.is_truthy();
```
MatthieuDartiailh, line 120: We can avoid playing with ref counts here:
```suggestion
    return PyObject_IsTrue( self->atomref );
```

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,335 @@
...
+    if( opid == Py_EQ )
+    {
+        if( SubscriptionObserver::TypeCheck( other ) )
+        {
+            SubscriptionObserver* so_other = reinterpret_cast<SubscriptionObserver*>( other );
+            cppy::ptr sref( cppy::incref( self->atomref ) );
+            cppy::ptr sname( cppy::incref( self->name ) );
+            cppy::ptr oref( cppy::incref( so_other->atomref ) );
+            cppy::ptr oname( cppy::incref( so_other->name ) );
+            if( sref.richcmp( oref, Py_EQ ) && sname.richcmp(oname, Py_EQ ))
```
MatthieuDartiailh, line 168: Same here
```suggestion
            if(
                PyObject_RichCompareBool( self->atomref, so_other->atomref, Py_EQ )
                && PyObject_RichCompareBool( self->name, so_other->name, Py_EQ )
            )
```

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,335 @@
...
+ *         engine.update(owner, self.name)
+ */
+PyObject*
+SubscriptionObserver_call( SubscriptionObserver* self, PyObject* args, PyObject* kwargs )
+{
+
+    cppy::ptr atomrefptr( cppy::incref( self->atomref ) );
+    if( atomrefptr.is_truthy() )
+    {
+        cppy::ptr owner( PyObject_CallNoArgs( atomrefptr.get() ) );
```
MatthieuDartiailh, line 139: Same here:
```suggestion
    if( PyObject_IsTrue( self->atomref ) )
    {
        cppy::ptr owner( PyObject_CallNoArgs( self->atomref ) );
```

MatthieuDartiailh (commented):
I made several suggestions to simplify your handling of reference counts. I tested locally and it works but feel free to give it another spin.

Could you also add a changelog entry ?

frmdstryr:
Thanks for the review.

I tried moving the whole standard_tracer to cpp [here](https://github.com/frmdstryr/enaml/tree/standard-tracer-cpp) but this branch actually faster somehow so I think it's better to just go with this branch as it's also much simpler.

MatthieuDartiailh:
Can you rebase on main to get the CI to pass ?

sccolbert:
I’d like to review the c++ code before this is merged.

MatthieuDartiailh:
Sure Chris I will wait for your approval before merging.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+    Py_CLEAR( self->atomref );
+    Py_CLEAR( self->name );
+}
+
+
+int
+SubscriptionObserver_traverse( SubscriptionObserver* self, visitproc visit, void* arg )
+{
+    Py_VISIT( self->atomref );
+    Py_VISIT( self->name );
```
sccolbert, line 88: This VISIT is not needed, since strings cannot be a part of cycles.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+    Py_VISIT(Py_TYPE(self));
```
sccolbert, line 89: This VISIT is also not needed. This type object will not be a part of GC cycles.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+ *      if engine is not None:
+ *         engine.update(owner, self.name)
+ */
+PyObject*
+SubscriptionObserver_call( SubscriptionObserver* self, PyObject* args, PyObject* kwargs )
+{
+
+    if( PyObject_IsTrue( self->atomref ) )
+    {
+        cppy::ptr owner( PyObject_CallNoArgs( self->atomref ) );
```
sccolbert, line 124: This need to be check for an exception thrown. After this line, you need:

```
if( !owner)
{ 
    return 0;
}
```

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+        cppy::ptr engine( owner.getattr("_d_engine") );
```
sccolbert, line 125: Same as above.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+        if ( !engine.is_none() )
+        {
+            cppy::ptr update_args( PyTuple_New( 2 ) );
+            if( !update_args )
+                return 0;
+            PyTuple_SET_ITEM( update_args.get(), 0, cppy::incref( owner.get() ) );
+            PyTuple_SET_ITEM( update_args.get(), 1, cppy::incref( self->name ) );
+            cppy::ptr update( engine.getattr( "update" ) );
```
sccolbert, line 133: Same as above.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+            return update.call( update_args );
+        }
+    }
+    Py_RETURN_NONE;
+}
+
+
+PyObject*
+SubscriptionObserver_richcompare( SubscriptionObserver* self, PyObject* other, int opid )
```
sccolbert, line 142: I don't think this richcompare method is needed.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+PyObject*
+SubscriptionObserver_get_ref( SubscriptionObserver* self, void* context )
+{
+    return cppy::incref( self->atomref );
+}
+
+
+PyObject*
+SubscriptionObserver_set_ref( SubscriptionObserver* self, PyObject* value, void* context )
```
sccolbert, line 182: I think the only time this is ever called to be set is with `None`. I think we can just check that `value == Py_NONE` here and throw an exception otherwise.

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+{
+    if( reinterpret_cast<PyObject*>( self ) == value )
+        return 0;
+    cppy::ptr old( self->atomref );
```
sccolbert, line 186: This line could be `cppy::clear( self->atomref )`;

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+    self->atomref = cppy::incref( value );
```
sccolbert, line 187: If the comment about `None` is correct, this can be `self->atomref = cppy::incref( PY_NONE )`

enaml/src/subscription_observer.cpp
```diff
@@ -0,0 +1,332 @@
...
+SubscriptionObserver_get_ref( SubscriptionObserver* self, void* context )
+{
+    return cppy::incref( self->atomref );
+}
+
+
+PyObject*
+SubscriptionObserver_set_ref( SubscriptionObserver* self, PyObject* value, void* context )
+{
+    if( reinterpret_cast<PyObject*>( self ) == value )
```
sccolbert, line 184: I don't think this is needed, but even if it were, you would need to set an exception before returning `0`.

sccolbert (changes requested):

enaml/src/subscription_observer.cpp
MatthieuDartiailh: It's been too long since I did C Pyrhon API stuff I should have seen this.

enaml/src/subscription_observer.cpp
frmdstryr: Agree, cppy:replace seems to do both in one go.

enaml/src/subscription_observer.cpp
frmdstryr: Doesn't the atom observerpool use richcmp internally? 

Is this [new test](https://github.com/nucleic/enaml/pull/562/files#diff-70fb0bd6bd7f8242e3158b1add0f44e80ebd5ba758f11ed8dbfc422b627f27cdR33-R37)  correct?

sccolbert: It does, but that feature is not needed for this `SubcriptionObserver`. A new observer is created every time the expression is evaluated, so the richcmp feature isn't necessary. It's better to short-circuit that comparison using just pointer identity.

https://github.com/nucleic/enaml/blob/main/enaml/core/standard_tracer.py#L112-L122

enaml/src/subscription_observer.cpp
frmdstryr: I did a search and it looks like atomlist misses these as well? https://github.com/nucleic/atom/blob/main/atom/src/atomlist.cpp#L788-L791

enaml/src/subscription_observer.cpp
sccolbert: We could potentially be more intelligent there and do a diff between the old and new dependencies, and only change observers as needed. But up til now, it hasn't been a bottleneck, and the current code is technically correct.

enaml/src/subscription_observer.cpp
sccolbert: That slipped by me, but needs to be corrected: @MatthieuDartiailh 
https://github.com/nucleic/atom/blame/main/atom/src/atomlist.cpp#L788-L791

tests/core/test_subscription_observer.py
```diff
@@ -34,7 +34,7 @@ def test_subscription_observer_compare():
     label = Label()
```
sccolbert, line 34: This whole test can go away. The only thing we care about is pointer identity, which is different for two different objects, by definition.

frmdstryr:
The original example now runs in 9 seconds, down from the original 50 seconds.

MatthieuDartiailh:
That's a really nice improvement !

sccolbert (approved):

PR #566, opened 2025-01-10
Add dynamicscope items, keys, values, update, and iter

### LINKED ISSUE #565: Errors cannot be printed/formatted in 3.12+ because DynamicScope object is not iterable
It seems any error in an enaml expression/event callback with python 3.12+ causes the application to sigabort because errors cannot be printed (by pyqt or any other logging).

For example
```python
import traceback
from atom.api import Atom, Str, Int, Instance, List


def log_error(e):
    traceback.print_exception(e)

enamldef Main(Window):
    Container:
        Label:
            text = "Hello"
        PushButton:
            text = "Click me"
            clicked ::
                try:
                    print(missing_var)
                except Exception as e:
                    log_error(e)
```

Causes this:

```python-traceback
  File "tests/error.enaml", line 16, in f
    try:
NameError: name 'missing_var' is not defined

During handling of the above exception, another exception occurred:

  File "/home/jrm/projects/enaml/enaml/qt/qt_abstract_button.py", line 83, in on_clicked
    self.declaration.clicked(checked)
  File "/home/jrm/projects/enaml/enaml/core/declarative_meta.py", line 72, in declarative_change_handler
    engine.write(owner, change['name'], change)
  File "/home/jrm/projects/enaml/enaml/core/expression_engine.py", line 210, in write
    pair.writer(owner, name, change)
  File "/home/jrm/projects/enaml/enaml/core/standard_handlers.py", line 82, in __call__
    call_func(func, (), {}, scope)
  File "tests/error.enaml", line 19, in f
    log_error(e)
  File "tests/error.enaml", line 7, in log_error
    traceback.print_exception(e)
  File "/home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/traceback.py", line 124, in print_exception
    te = TracebackException(type(value), value, tb, limit=limit, compact=True)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/traceback.py", line 767, in __init__
    suggestion = _compute_suggestion_error(exc_value, exc_traceback, wrong_name)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/traceback.py", line 1091, in _compute_suggestion_error
    list(frame.f_locals)
TypeError: 'enaml.dynamicscope.DynamicScope' object is not iterable
Aborted (core dumped)
```

Without the  try/except it occurs in pyqt. 

```
Thread 1 "python" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
warning: 44     ./nptl/pthread_kill.c: No such file or directory
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7c4526e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7c288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff5cbd012 in ?? () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/Qt6/lib/libQt6Core.so.6
#6  0x00007ffff5cfaa45 in ?? () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/Qt6/lib/libQt6Core.so.6
#7  0x00007ffff5cbe945 in QMessageLogger::fatal(char const*, ...) const () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/Qt6/lib/libQt6Core.so.6
#8  0x00007ffff66456de in pyqt6_err_print() () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/QtCore.abi3.so
#9  0x00007ffff664e2cd in PyQtSlotProxy::unislot(void**) () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/QtCore.abi3.so
#10 0x00007ffff664f307 in PyQtSlotProxy::qt_metacall(QMetaObject::Call, int, void**) () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/QtCore.abi3.so
#11 0x00007ffff5dde36c in ?? () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/Qt6/lib/libQt6Core.so.6
#12 0x00007ffff14b9576 in QAbstractButton::clicked(bool) () from /home/jrm/.local/share/mamba/envs/enaml/lib/python3.12/site-packages/PyQt6/Qt6/lib/libQt6Widgets.so.6
```

In enaml-web any error now shuts down the whole application..

frmdstryr:
This at least fixes #565 and makes ipython embed work in an expression/event handler.

I don't think doing the full scope (walking parents and getting members) is necessary but I can try that if needed.

### DIFF
--- a/enaml/src/dynamicscope.cpp
+++ b/enaml/src/dynamicscope.cpp
@@ -784,9 +784,105 @@ DynamicScope_contains( DynamicScope* self, PyObject* key )
     return test_dynamic_attr( self->owner, key );
 }
 
+PyObject*
+_DynamicScope_dict( DynamicScope* self)
+{
+    cppy::ptr items( PyDict_New() );
+    if ( !items )
+        return 0;
+
+    if ( PyDict_Update( items.get(), self->f_builtins ) )
+        return 0;
+
+    if ( PyDict_Update( items.get(), self->f_globals ) )
+        return 0;
+
+    if ( PyDict_Update( items.get(), self->f_locals ) )
+        return 0;
+
+    if ( self->f_writes && PyDict_Update( items.get(), self->f_writes ) )
+        return 0;
+
+    if ( self->change && PyDict_SetItemString( items.get(), "change", cppy::incref(self->change) ) )
+        return 0;
+
+    if ( PyDict_SetItemString( items.get(), "self", cppy::incref(self->owner) ) )
+        return 0;
+
+    return items.release();
+}
+
+
+PyObject*
+DynamicScope_items( DynamicScope* self)
+{
+    cppy::ptr dict( _DynamicScope_dict( self ) );
+    if ( !dict )
+        return 0;
+    return PyDict_Items( dict.get() );
+}
+
+
+PyObject*
+DynamicScope_keys( DynamicScope* self)
+{
+    cppy::ptr dict( _DynamicScope_dict( self ) );
+    if ( !dict )
+        return 0;
+    return PyDict_Keys( dict.get() );
+}
+
+
+PyObject*
+DynamicScope_values( DynamicScope* self)
+{
+    cppy::ptr dict( _DynamicScope_dict( self ) );
+    if ( !dict )
+        return 0;
+    return PyDict_Values( dict.get() );
+}
+
+
+PyObject*
+DynamicScope_iter( DynamicScope* self)
+{
+    cppy::ptr dict( _DynamicScope_dict( self ) );
+    if ( !dict )
+        return 0;
+    return dict.iter();
+}
+
+
+PyObject*
+DynamicScope_update( DynamicScope* self, PyObject* arg )
+{
+    if( !PyMapping_Check( arg ) )
+        return cppy::type_error( arg, "mapping" );
+
+    cppy::ptr items( cppy::incref( arg ) );
+    cppy::ptr iter( items.iter() );
+    if ( !iter )
+        return 0;
+
+    cppy::ptr key;
+    while ( (key = iter.next())  )
+    {
+        cppy::ptr value ( PyObject_GetItem( items.get(), key.get() ) );
+        if ( !value )
+            return 0;
+        if ( DynamicScope_setitem( self, key.get(), value.get() ) )
+            return 0;
+    }
+
+    Py_RETURN_NONE;
+}
 
 static PyMethodDef DynamicScope_methods[] = {
     {"get",    reinterpret_cast<PyCFunction>(DynamicScope_get), METH_VARARGS, ""},
+    {"items",  reinterpret_cast<PyCFunction>(DynamicScope_items), METH_NOARGS, ""},
+    {"keys",   reinterpret_cast<PyCFunction>(DynamicScope_keys), METH_NOARGS, ""},
+    {"values", reinterpret_cast<PyCFunction>(DynamicScope_values), METH_NOARGS, ""},
+    {"update", reinterpret_cast<PyCFunction>(DynamicScope_update), METH_O, ""},
     { 0 }  // Sentinel
 };
 
@@ -801,6 +897,7 @@ static PyType_Slot DynamicScope_Type_slots[] = {
     { Py_mp_subscript, void_cast( DynamicScope_getitem ) },      /* mp_subscript */
     { Py_mp_ass_subscript, void_cast( DynamicScope_setitem ) },  /* mp_ass_subscript */
     { Py_sq_contains, void_cast( DynamicScope_contains ) },      /* sq_contains */
+    { Py_tp_iter, void_cast( DynamicScope_iter ) },              /* tp_iter */
     { 0, 0 },
 };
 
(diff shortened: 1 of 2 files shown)


sccolbert:
Before we go this route, I'd like it if someone investigated the change between Py2 and Py3 to see if that really is the problem. It may be (but probably unlikely) that everyone assumes `f_locals` is a `dict` when that's really not the case. We *really* need to understand what protocol is expected of `f_locals` by the interpreter. Because if the interpreter is now expecting a proper `dict` actual, then we are in trouble, and will need to re-think all this magic we are doing.

frmdstryr:
Both [frame.f_locals](https://docs.python.org/3/reference/datamodel.html#frame.f_locals) and [PyFrame_GetLocals](https://docs.python.org/3/c-api/frame.html#c.PyFrame_GetLocals) say mapping.

The traceback module uses [`list(frame.f_locals)`](https://github.com/python/cpython/blob/main/Lib/traceback.py#L1507) and [`locals.items()`](https://github.com/python/cpython/blob/main/Lib/traceback.py#L315-L316) which should be part of mapping at least according to [collections.abc](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes).

For reasons I don't understand the Dynamicscope has [Py_TPFLAGS_DICT_SUBCLASS](https://github.com/nucleic/enaml/blob/main/enaml/src/dynamicscope.cpp#L820) set but not Py_TPFLAGS_MAPPING so it is a dict according to PyDict_Check. Idk how it passes the PyMapping_Check yet.

Either way it at least needs to implement the `tp_iter`?

sccolbert:
All need to spend some time tracking this down to make sure we do the right
thing, but here is an issue to start your research if you wanna do it in
parallel with me:
https://bugs.python.org/issue5945

sccolbert:
It looks like `PyMapping_Check` is deficient in actually fully checking for the mapping protocol. You can see what it checks here:
https://github.com/python/cpython/blob/5e65a1acc0b630397f1d190aed279114e6e99612/Include/cpython/object.h#L125
https://github.com/python/cpython/blob/5e65a1acc0b630397f1d190aed279114e6e99612/Objects/abstract.c#L2257

But then the mapping functions go on to call other methods that should exist on the object (just read thru the rest of the functions below `PyMapping_Check`.

frmdstryr:
Agree on PyMapping_Check. I looked through [bytecodes.c](https://github.com/python/cpython/blob/main/Python/bytecodes.c) for uses of locals and it seems to only use PyObject_SetItem, PyObject_GetItemOptional, and PyObject_DelItem. 

So what do you think should be done?

MatthieuDartiailh:
For each of the added methods, I would like a clear view of when they are expected to be used and why we are fine paying the price of creating a dict.
Regarding the traversal of parents, I believe it could be valuable as otherwise figuring out the names accessible from parents can be quite tough.

sccolbert:
To do this properly, and without over-allocating upfront, we'll need to produce a proper iterator over the dynamic scope.

Likely the easiest way to do this is to keep a `_DynamicScope` class in `c++` which implements the high-perf `__getitem__` methods and friends. Then subclass that class in Python to implement the rest of the mapping protocol. You can make the `f_locals` `f_globals` `f_builtins` etc... as readonly attributes on the base class, then access those attributes from the subclass to implement `key` `items` `iter` etc. and leverage `itertools.chain` to cover everything.

Those python methods don't need to be fast, they just need to be present.

sccolbert:
But one other thing to keep in mind is that `DynamicScope` has behavior that supports the tracing machinery, so if you expose `DynamicScope` to python code, it could be a foot-gun where something isn't traced properly.

I would also be fine with `DynamicScope` just returning an empty iterator for all of the iterator/dict protocol stuff.

MatthieuDartiailh:
@sccolbert do you have any specific concern regarding tracing ? I do not really see what could happen from exposing interfaces that do not allow to mutate the scope.
Personally, I have two concerns:
- not providing anything makes the debugging experience worse and since we already suffer from zero IDE integration, I find it sad.
- providing these methods and matching the semantic expectations people may have (items reflect the actual dict content even is the dict is mutated) is a huge work.
A frozen items like in Python 2 would probably be sufficient for debugging purposes and hopefully the limitations won't be noticed.

PR #567, opened 2025-01-16
Update funchelper call_func

frmdstryr:
All of the argument types are already checked so it is safe to use the faster variant.

### DIFF
--- a/enaml/src/funchelper.cpp
+++ b/enaml/src/funchelper.cpp
@@ -20,17 +20,18 @@ from Python's funcobject.c
 
 */
 PyObject*
-call_func( PyObject* mod, PyObject* args )
+call_func( PyObject* mod, PyObject *const *args, Py_ssize_t nargs )
 {
-    PyObject* func;
-    PyObject* func_args;
-    PyObject* func_kwargs;
-    PyObject* func_locals = Py_None;
-
-    if( !PyArg_UnpackTuple( args, "call_func", 3, 4, &func, &func_args, &func_kwargs, &func_locals ) )
+    if( !(nargs == 3 || nargs == 4) )
     {
+        PyErr_SetString( PyExc_TypeError, "call_func must have 3 or 4 arguments" );
         return 0;
     }
+
+    PyObject* func = args[0];
+    PyObject* func_args = args[1];
+    PyObject* func_kwargs = args[2];
+    PyObject* func_locals = nargs == 4 ? args[3] : Py_None;
 
     if( !PyFunction_Check( func ) )
     {
@@ -115,7 +116,7 @@ funchelper_modexec( PyObject *mod )
 
 static PyMethodDef
 funchelper_methods[] = {
-    { "call_func", ( PyCFunction )call_func, METH_VARARGS,
+    { "call_func", ( PyCFunction )call_func, METH_FASTCALL,
       "call_func(func, args, kwargs[, locals])" },
     { 0 } // sentinel
 };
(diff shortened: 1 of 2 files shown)


frmdstryr:
I added a benchmark test on my computer. It's a bit hard to get accurate results due to the cpu throttling (whichever test runs first is slower). 

Test

```python
@pytest.mark.skipif(not BENCHMARK_INSTALLED, reason="benchmark is not installed")
@pytest.mark.benchmark(group="call_func")
def test_calling_function_benchmark(benchmark):
    """Test calling a function with locals.

    """
    benchmark.pedantic(call_func, args=(func, (), {}, {'a': 1}), iterations=1000, rounds=100000, warmup_rounds=100)
```
If I run the fastcall variant first these are the results

```
----------------------------------------------------------------------------------------- benchmark 'call_func': 2 tests -----------------------------------------------------------------------------------------
Name (time in ns)                            Min                 Max                Mean             StdDev              Median               IQR             Outliers  OPS (Mops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_calling_function_benchmark          92.6650 (1.0)      501.9160 (1.20)     113.8869 (1.0)       6.2469 (1.0)      113.5360 (1.0)      3.4460 (1.08)     4351;3667        8.7806 (1.0)      100000        1000
test_calling_function_benchmark_old     107.6770 (1.16)     418.9320 (1.0)      127.2937 (1.12)     18.7238 (3.00)     122.8500 (1.08)     3.1820 (1.0)      4840;6832        7.8558 (0.89)     100000        1000
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```


If I run the fastcall varant second these are the results

```
----------------------------------------------------------------------------------------- benchmark 'call_func': 2 tests -----------------------------------------------------------------------------------------
Name (time in ns)                            Min                 Max                Mean             StdDev              Median               IQR             Outliers  OPS (Mops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_calling_function_benchmark          94.0200 (1.0)      365.8000 (1.0)      119.1095 (1.0)      18.3048 (4.54)     114.8490 (1.0)      3.6615 (1.02)     5175;6754        8.3956 (1.0)      100000        1000
test_calling_function_benchmark_old     114.9570 (1.22)     464.4490 (1.27)     122.9618 (1.03)      4.0326 (1.0)      122.6720 (1.07)     3.5820 (1.0)      9027;1986        8.1326 (0.97)     100000        1000
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

enaml/src/funchelper.cpp
```diff
@@ -20,18 +20,19 @@ from Python's funcobject.c
...
-    PyObject* func_locals = Py_None;
-
-    if( !PyArg_UnpackTuple( args, "call_func", 3, 4, &func, &func_args, &func_kwargs, &func_locals ) )
+    if( !(nargs == 4 || nargs == 3) )
     {
+        PyErr_SetString( PyExc_TypeError, "call_func must have 3 or 4 arguments" );
         return 0;
     }
 
+    PyObject* func = args[0];
```
sccolbert, line 31: This is not the proper way to access the items in a tuple.

You should use `PyTuple_GET_ITEM`, which returns a borrowed reference.

frmdstryr: I thought the point of using [METH_FASTCALL](https://docs.python.org/3/c-api/structures.html#c.PyCFunctionFast) was to avoid packing/unpacking tuples?

sccolbert: I stand corrected. Hadn't seen that convention before. Sorry for the noise!

MatthieuDartiailh (approved):

PR #569, opened 2025-01-20
Refactor declarative function and make call_func kwargs optional

frmdstryr:
This refactors declarative_function's Invoke to:
- use vectorcalls 
- cache the strings used
- Avoid creating empty f_locals dict 
- Avoid creating empty kwargs dict

It also slightly modifies call_func:
- require 4 arguments because I can't find any uses with 3 arguments
- allow func_kwargs to be null or None (see below). 
 
Since all standard handlers create an empty dict and pass it to kwargs that just gets checked for a dict then ignored because it is empty.  Likewise previously every dfunc call that has no kwargs was internally creating a dict just to do nothing with it in call_func. With this change the DFunc__call__'s PyObject* kwargs can be passed to call_func untouched (and if null does nothing).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 *.un~
 *.swp
+*.kate-swp
 *.egg-info
 *.so
 *.pyd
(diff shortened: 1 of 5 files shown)


frmdstryr:
After profiling this has practically no impact so forget it.

PR #571, opened 2025-01-29
Add a proper dynamicscope iterator

frmdstryr:
It is not very efficient to the use of dir() and a set to track used keys... but makes python 3.12+ provide useful messages for name errors and attribute errors. 

I'd like to add more tests after any initial feedback.

Eg:

```python
    Container:
        PushButton:
            text = "Click me!"
            clicked :: print(windo.size())
```
Gives `NameError: name 'windo' is not defined. Did you mean: 'window'?`

From looking at the traceback source code it ignores giving suggestions if there are > 750 items (https://github.com/python/cpython/blob/main/Lib/traceback.py#L1465). Which I can see it easily exceeding.. So it might be good to also skip globals and builtins?

### DIFF
--- a/enaml/core/dynamicscope.py
+++ b/enaml/core/dynamicscope.py
@@ -1,0 +1,107 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2025, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ------------------------------------------------------------------------------
+from functools import partial
+from itertools import chain, repeat
+from enaml.core.declarative import Declarative
+from enaml.core._dynamicscope import _DynamicScope, UserKeyError
+
+
+def d_iter(owner: Declarative):
+    """Iterate attribute names of the declarative and it's ancestors.
+
+    Parameters
+    ----------
+    owner: Declarative
+        The declarative to walk.
+
+    Yields
+    ------
+    name: string
+        The attribute name
+
+    """
+    while owner is not None:
+        for name in dir(owner):
+            yield name
+        owner = owner._parent
+
+
+def include_key(key: str, used: set) -> bool:
+    """Filter function to determine whether the key should be included in the
+    dynamicscope's iter results.
+
+    Parameters
+    ----------
+    key: string
+        The scope key.
+    used: set[str]
+        The set of keys already seen.
+
+    Returns
+    -------
+    result: bool
+        Whether the key should be included.
+    """
+    if key.startswith("_") or key in used:
+        return False
+    used.add(key)
+    return True
+
+
+class DynamicScope(_DynamicScope):
+    """_DynamicScope is a C++ class which exposes the following attributes:
+
+    _owner
+    _change
+    _f_writes
+    _f_locals
+    _f_globals
+    _f_builtins
+
+    """
+
+    def __iter__(self):
+        """Iterate the keys available in the dynamicscope."""
+        used = set()
+        fwrites_it = iter(self._f_writes or ())
+        self_it = repeat("self", 1)
+        change_it = repeat("change", 1 if self._change else 0)
+        flocals_it = iter(self._f_locals)
+        fglobals_it = iter(self._f_globals)
+        fbuiltins_it = iter(self._f_builtins)
+        fields_it = d_iter(self._owner)
+        unique_scope_keys = partial(include_key, used=used)
+        return filter(
+            unique_scope_keys,
+            chain(
+                fwrites_it,
+                self_it,
+                change_it,
+                flocals_it,
+                fglobals_it,
+                fbuiltins_it,
+                fields_it,
+            )
+        )
+
+    def keys(self):
+        """Iterate the keys available in the dynamicscope."""
+        return iter(self)
+
+    def values(self):
+        """Iterate the values available in the dynamicscope."""
+        return (self[key] for key in self)
+
+    def items(self):
+        """Iterate the (key, value) pairs available in the dynamicscope."""
+        return ((key, self[key]) for key in self)
+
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
Do we agree this supersedes #566 ?

frmdstryr:
Yes, should I close that or push this on top of that branch?

sccolbert:
I think most of this should be done in a Python subclass of `DynamicScope`? It doesn't need to be fast, just correct. It's much easier to do that from Python. For example:

```python
from itertools import chain, repeat
from _dynamicscope import _DynamicScope


"""
_DynamicScope is a C++ class which exposes the following attributes:

_f_writes
_self_str
_change_str
_f_locals
_f_globals
_f_builtins

"""

class DynamicScope(_DynamicScope):

	def __iter__(self):
		fwrites_it = iter(self._f_writes)
		self_it = repeat(self._self_str, 1)
		change_it = repeat(self._change_str, 1)
		flocals_it = iter(self._f_locals)
		fglobals_it = iter(self._f_globals)
		fbuiltins_it = iter(self._f_builtins)
		parent_it = iter(self._parent)

		return chain(
			fwrites_it,
			self_it,
			change_it,
			flocals_it,
			fglobals_it,
			fbuiltins_it,
			parent_it
		)

	def keys(self):
		return iter(self)

	def values(self):
		return (self[key] for key in self)

	def items(self):
		return ((key, self[key]) for key in self)

	def update(self, items):
		for (key, value) in items:
			self[key] = value
```

frmdstryr:
Is it ok to expose the internal fields as properties? I guess it'd be fine with only getters.

sccolbert:
Yes. Make them read-only properties with a leading underscore. i.e. they
are "just an implementation detail" that the public shouldn't mess with.

enaml/core/dynamicscope.py
```diff
@@ -0,0 +1,58 @@
...
+            fglobals_it,
+            fbuiltins_it,
+            parent_it,
+        ))
+
+    def keys(self):
+        return iter(self)
+
+    def values(self):
+        return tuple(self[key] for key in self)
```
sccolbert, line 51: can we not return a generator expression for this method and `items`, or is it only `keys` that can return an iterator?

enaml/core/dynamicscope.py
```diff
@@ -0,0 +1,58 @@
...
+    def __iter__(self):
+        fwrites_it = iter(self._f_writes or ())
+        self_it = repeat("self", 1)
+        change_it = repeat("change", 1 if self._change else 0)
+        flocals_it = iter(self._f_locals)
+        fglobals_it = iter(self._f_globals)
+        fbuiltins_it = iter(self._f_builtins)
+        fields = set()
+        parent = self._owner
+        while parent is not None:
```
sccolbert, line 33: Why traverse the parent tree here ahead-of-time instead of allowing `chain` to naturally recurse into the parent iterator as needed?

enaml/core/dynamicscope.py
frmdstryr: Good point, not sure why I changed that.

enaml/core/dynamicscope.py
frmdstryr: Not sure I follow... the owner is a Declarative, the parent is another Declarative neither are iterable. It could use [traverse_ancestors](https://github.com/nucleic/enaml/blob/main/enaml/core/object.py#L330)?

Since the traceback module uses `list(frame.f_locals)`, if it doesn't remove duplicates it will likely quickly overflow the 750 item limit and not provide a suggestion at all...

frmdstryr: Or is there a way to get the parent dynamic scope? I don't see it anywhere from a quick look.

sccolbert: Sorry, yeah, I had it in my head that `parent` was another `DynamicScope` object. That said, I think we still put all of that logic into a generator function, instead of computing ahead-of-time.

sccolbert: For the 750 item limit, that really should be an issued raised with the core traceback module. Instead of 100% punting at 751 names, they should truncate their search space to the first 750 items, and just ignore the rest.

enaml/core/dynamicscope.py
```diff
@@ -0,0 +1,58 @@
...
+    def keys(self):
+        return iter(self)
+
+    def values(self):
+        return (self[key] for key in self)
+
+    def items(self):
+        return ((key, self[key]) for key in self)
+
+    def update(self, state):
```
sccolbert, line 56: Do we *need* an `update` method? All the rest of the methods are readonly. I know the C++ class implements `__setitem__`, but IIRC that's in order to handle local assignments. I'm not sure we really need this method.

frmdstryr: Technically no... I added it to support ipython embedding since it attempts to update the scope when exiting an embedded console.

enaml/core/dynamicscope.py
```diff
@@ -0,0 +1,107 @@
...
+            unique_scope_keys,
+            chain(
+                fwrites_it,
+                self_it,
+                change_it,
+                flocals_it,
+                fglobals_it,
+                fbuiltins_it,
+                fields_it,
+            ),
```
sccolbert, line 89: trailing comma?

enaml/core/dynamicscope.py
```diff
@@ -0,0 +1,107 @@
...
+        The scope key.
+    used: set[str]
+        The set of keys already seen.
+
+    Returns
+    -------
+    result: bool
+        Whether the key should be included.
+    """
+    if key.startswith("__") or key in used:
```
sccolbert, line 50: Should filter all keys with a single leading `_` instead of just double? i.e. only yield the "public" keys.

PR #572, opened 2025-02-01
Use yield from instead of chained iterables

sccolbert:
The `yield from` feature of Python 3 makes this cleaner.

I didn't test locally. Let's see what the CI tests show.


sccolbert:
@MatthieuDartiailh, @frmdstryr  this is ready to go if y'all approve.

MatthieuDartiailh (approved):

PR #574, opened 2025-03-08
import_hooks: Use cache_tag from PEP 421

AlexanderS:
Using the cache_tag from sys.implementation over manually creating a similar tag has the advantage, that it can correctly distinguishes between different python implementations.


MatthieuDartiailh:
Thanks for the PR. Since Enaml is strongly tied to CPython since it relies on some bytecode customization, I am not sure if it could work on other implementations but it makes sense to use the cache_tag since it is what it is meant for.

PR #609, opened 2026-03-31
Update Alias ext

frmdstryr:
- Adds decref of typeobj 
- Makes storage_str use interned
- Changes first arg (target) to only accept a str

The key, and items in the chain should be probably also be type checked to ensure it is ok to not use GC?


MatthieuDartiailh (approved):
This is already much better. Let's merge as is.

PR #612, opened 2026-04-01
Update Color ext

frmdstryr:


bri-tong (commented):
## Code Review — PR #612: Update Color ext

### Summary

This PR correctly fixes the `Color_dealloc` function in `colorext.cpp` to follow the proper CPython heap-type deallocation protocol: saving `Py_TYPE(self)` before freeing the object, then calling `Py_DECREF(tp)` afterward. This is the right pattern for types created via `PyType_FromSpec`, since they are heap-allocated and hold a refcount that must be managed. The fix is small, correct, and consistent with the pattern already applied in `alias.cpp` and `subscription_observer.cpp`.

### Blocking Issues

None. The change is correct.

### Non-blocking Suggestions

**The same bug exists in several other extension files.** Since this repo creates all its types via `PyType_FromSpec` (making them heap types), every `tp_dealloc` function should use the same save-type-then-decref pattern. The following files still use the old, buggy `Py_TYPE(self)->tp_free(...)` without `Py_DECREF`:

| File | Dealloc function(s) |
|---|---|
| `enaml/src/fontext.cpp` | `Font_dealloc` |
| `enaml/src/callableref.cpp` | `CallableRef_dealloc` |
| `enaml/src/declarative_function.cpp` | `DFunc_dealloc`, `BoundDMethod_dealloc` |
| `enaml/src/dynamicscope.cpp` | `Nonlocals_dealloc`, `DynamicScope_dealloc` |
| `enaml/src/signaling.cpp` | `Signal_dealloc`, `_Disconnector_dealloc`, `BoundSignal_dealloc` |
| `enaml/src/weakmethod.cpp` | `WeakMethod_dealloc` |

Consider fixing all of these in this PR (or a follow-up) for consistency and to avoid type refcount leaks across the board. Without the `Py_DECREF(tp)`, the type object will never be freed, which is a memory leak — albeit typically minor since types are usually long-lived singletons.

### Questions / Clarifications

None — the intent is clear and the implementation is correct.

MatthieuDartiailh (approved):

PR #613, opened 2026-04-07
Update declarative function

frmdstryr:
Addresses recent issues and updates to use vectorcalls. 

A simple timeit calling a dfunc shows about a 30% speedup with this change (but still over order of magnitude slower than a normal py fn call). 

I think it is safe to assume the caller "owns" a reference to several of the arguments so this removes a few increfs (pfunc, pself, and argsptr).

### DIFF
--- a/enaml/src/declarative_function.cpp
+++ b/enaml/src/declarative_function.cpp
@@ -22,6 +22,13 @@ namespace enaml
 {
 
 
+static PyObject* builtins_str;
+static PyObject* d_storage_str;
+static PyObject* get_str;
+static PyObject* globals_str;
+static PyObject* super_str;
+
+
 // POD struct - all member fields are considered private
 struct DFunc
 {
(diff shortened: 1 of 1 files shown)


frmdstryr:
It looks like call_func has to generate a new function with every call. I wonder if the bound dfunc could somehow cache it

enaml/src/declarative_function.cpp
```diff
@@ -480,18 +503,9 @@ bool BoundDMethod::Ready()
...
-    {
-        pymethod = pyobject_cast( freelist[ --numfree ] );
-        _Py_NewReference( pymethod );
-    }
-    else
-    {
-        pymethod = PyType_GenericAlloc( BoundDMethod::TypeObject, 0 );
-        if( !pymethod )
-            return 0;
-    }
```
MatthieuDartiailh, line 494: Why did you remove the freelist ?

MatthieuDartiailh:
> It looks like call_func has to generate a new function with every call. I wonder if the bound dfunc could somehow cache it

Could you point me to this I do not remember why it is so ?

MatthieuDartiailh (commented):
One question

enaml/src/declarative_function.cpp
frmdstryr: Because the first issue said there was issues with GC and suggested doing that and I didn't want to look into it. 

But since you asked... I re-added it after digging through methodobject.c and the freelist code that I think simply adding the PyObject_GC_Track fixes the GC issue. The BoundDMethod::TypeObject will still leak references (one per freelist obj) but cant be subclassed so I think that's fine.

MatthieuDartiailh (approved):

frmdstryr:
> > It looks like call_func has to generate a new function with every call. I wonder if the bound dfunc could somehow cache it
> 
> Could you point me to this I do not remember why it is so ?

Looking at the source of [PyEval_EvalCodeEx](https://github.com/python/cpython/blob/main/Python/ceval.c#L2129) it creates a new function using _PyFunction_FromConstructor mostly just copying the arguments. 

I'm not sure if it can avoid this and somehow just use _PyEval_Vector which takes the locals. I'll open a PR if I can get something working.

frmdstryr:
I noticed PyFunctionObject has it's own func_globals and func_builtins fields (see https://github.com/python/cpython/blob/main/Objects/funcobject.c#L209-L210).  Do you see any problems with casting and using those directly (instead of doing getattrs)?

I was able to get func calls to ~3x slower than normal py fn calls (down from 9x slower in current master) using this branch https://github.com/nucleic/enaml/compare/main...frmdstryr:enaml:improve-speed-callfunc but that requires reworking some more things.

MatthieuDartiailh:
I see no reason why not using the existing fields. I had no time to check your branch but I will try to.

PR #614, opened 2026-04-07
Update dynamicscope

frmdstryr:
This addresses the recent issues, changes run_tracer and Dynamicscope_get to use a vectorcall, and makes all the magics use interned strings. 

In a simple benchmark (updating a local and global counter within a clicked block) I see about a 10% speedup.

### DIFF
--- a/enaml/src/dynamicscope.cpp
+++ b/enaml/src/dynamicscope.cpp
@@ -142,16 +142,8 @@ maybe_translate_key_error()
 inline bool
 run_tracer( PyObject* tracer, PyObject* owner, PyObject* name, PyObject* value )
 {
-    cppy::ptr handler( PyObject_GetAttr( tracer, dynamic_load_str ) );
-    if( !handler )
-        return false;
-    cppy::ptr args( PyTuple_New( 3 ) );
-    if( !args )
-        return false;
-    PyTuple_SET_ITEM( args.get(), 0, cppy::incref( owner ) );
-    PyTuple_SET_ITEM( args.get(), 1, cppy::incref( name ) );
-    PyTuple_SET_ITEM( args.get(), 2, cppy::incref( value ) );
-    cppy::ptr res( handler.call( args ) );
+    PyObject* args[] = {tracer, owner, name, value};
+    cppy::ptr res( PyObject_VectorcallMethod( dynamic_load_str, args, 4 | PY_VECTORCALL_ARGUMENTS_OFFSET, 0) );
     if( !res )
         return false;
     return true;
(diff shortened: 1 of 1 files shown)


enaml/src/dynamicscope.cpp
```diff
@@ -743,23 +727,23 @@ DynamicScope_contains( DynamicScope* self, PyObject* key )
         return 1;
 
     // 'self' magic
-    if( strcmp( (char *)PyUnicode_AsUTF8( key ), "self" ) == 0 )
+    if( key == self_str )
```
MatthieuDartiailh: Out of curiosity did you test manually checking from Python ? Does the fact you interned once guarantees the pointer equality ?

MatthieuDartiailh (commented):
Looks nice. I have one question around str comparison.

enaml/src/dynamicscope.cpp
frmdstryr: My understanding is that if both strings are interned then it is guaranteed to have pointer equality (see https://github.com/python/cpython/blob/main/InternalDocs/string_interning.md#string-interning).

Even though the `key` in dynamicscope is not forced to be interned in practice all of these strings should always interned (except the tracer string) because from looking at cpython source it appears they intern any constants strings used in code objects that match the identifier pattern `[a-zA-Z0-9_]` (see https://github.com/python/cpython/blob/main/Objects/codeobject.c#L117C1-L117C21). I had originally renamed the `_[tracer]` to `__tracer__` and that worked without needing to explicitly intern it.

The references held effectively make them immortal so they shouldn't ever change so I think this is safe to do this. Unless you can think of anything else that would break it?

sccolbert:
The `_[tracer]` form was chosen explicitly because it is not a valid
python identifier. It makes it harder to access the tracer within
expressions and prevents a potential footgun since it also can't collide
with any other valid python identifier.

frmdstryr:
I made it force the key to be interned by copying the [sys_intern_impl](https://github.com/python/cpython/blob/main/Python/sysmodule.c#L1004) (also see [PyUnicode_InternInPlace](https://github.com/python/cpython/blob/main/Objects/unicodeobject.c#L14369C1-L14369C24)). It still is ~10% faster than the main branch but now does not depend on the key being interned.

MatthieuDartiailh:
I am a bit concerned by the idea of interning all names looked up in the dynamic scope. It could blow up memory. I need to think more about it.

frmdstryr:
It looks like PyObject_SetAttr interns always interns the attr and the default getattr  (_PyObject_GenericGetAttrWithDict) uses ptr comparisons (at least for descriptors in _PyType_LookupRef), but anyways IDK so I reverted it. It's only like a 5% difference and probably negligible in most code anyways.

Edit: Another approach is to save the hashes and use that. It is slightly faster than both https://github.com/frmdstryr/enaml/tree/update-dscope-hash .

MatthieuDartiailh (approved):

PR #618, opened 2026-04-21
Drop dead PY311 branches missed by copilot

frmdstryr:
Due to the min version, PY311 is now constant `True` and can be removed.

### DIFF
--- a/enaml/compat.py
+++ b/enaml/compat.py
@@ -10,8 +10,6 @@ import re
 import codecs
 import tokenize
 
-PY311 = sys.version_info >= (3, 11)
-
 PY312 = sys.version_info >= (3, 12)
 
 PY313 = sys.version_info >= (3, 13)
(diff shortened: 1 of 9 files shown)


frmdstryr:
It clean things up if the `load_helper` could somehow manage pushing the null value into the correct spot.

MatthieuDartiailh (approved):
Very nice

MatthieuDartiailh:
> It clean things up if the `load_helper` could somehow manage pushing the null value into the correct spot.

Could you point me exactly to what you mean ?

frmdstryr:
If you search for load_helper there's several instances of code like:

```py
        # Python 3.11 and 3.12 requires a NULL before a function that is not a method
        # Python 3.13 one after
        if not PY313 and PY311:
            cg.push_null()
        load_helper(cg, 'add_alias')
        if PY313:
            cg.push_null()
```

I was wondering if load_helper could handle putting the push_null at the correct spot dependent on the version, but I couldn't find a clean way to do it because it depends on the number of variables.

I did it for make_function here https://github.com/nucleic/enaml/commit/7478234c799a6ab3aaa631cafa66f564496afb2e

MatthieuDartiailh:
Thanks for the clarification. I will merge.

Issue #617, opened 2026-04-20
lambda kwarg default not getting captured on 3.13+


frmdstryr:
The following works on 3.12 and lower but on 3.13 and 3.14 it gives an error saying m is missing despite a default being provided.

```python
from enaml.widgets.api import Window, Container, Label


enamldef Main(Window): window:
    attr mode: str = "a"
    attr keymap = {
        # Use case is populating from an Enum's items
        k: lambda m=k: setattr(self, "mode", m) for k in ("a", "b")
    }

    func on_key_released(event):
        if handler := keymap.get(event.text()):
            handler()

    activated :: self.proxy.widget.keyReleaseEvent = self.on_key_released

    Container:
        Label:
            text << f"Mode is: {mode}"
        Label:
            text = "Push a or b to change mode"
```

Edit: The operator code is
```py
Sanitize op:
<RESUME arg=0 location=InstrLocation(lineno=0, end_lineno=1, col_offset=0, end_col_offset=0)>
<LOAD_CONST arg=('a', 'b') location=InstrLocation(lineno=8, end_lineno=8, col_offset=57, end_col_offset=67)>
<GET_ITER location=InstrLocation(lineno=8, end_lineno=8, col_offset=57, end_col_offset=67)>
<LOAD_FAST_AND_CLEAR arg='k' location=InstrLocation(lineno=6, end_lineno=9, col_offset=18, end_col_offset=5)>
<SWAP arg=2 location=InstrLocation(lineno=6, end_lineno=9, col_offset=18, end_col_offset=5)>
<bytecode.instr.TryBegin object at 0x7867cef26640>
<BUILD_MAP arg=0 location=InstrLocation(lineno=6, end_lineno=9, col_offset=18, end_col_offset=5)>
<SWAP arg=2 location=InstrLocation(lineno=6, end_lineno=9, col_offset=18, end_col_offset=5)>
<bytecode.instr.Label object at 0x7867cf793520>
<FOR_ITER arg=<bytecode.instr.Label object at 0x7867cf793d80> location=InstrLocation(lineno=8, end_lineno=8, col_offset=57, end_col_offset=67)>
<STORE_FAST_LOAD_FAST arg=('k', 'k') location=InstrLocation(lineno=8, end_lineno=8, col_offset=52, end_col_offset=53)>
<LOAD_FAST_BORROW arg='k' location=InstrLocation(lineno=8, end_lineno=8, col_offset=20, end_col_offset=21)>
[...]

PR #620, opened 2026-04-22
Execute SET_FUNCTION_ATTRIBUTE before wrapping for #617

frmdstryr:
On 3.13+, when a function has defaults it uses the new [SET_FUNCTION_ATTRIBUTE](https://docs.python.org/3/library/dis.html#opcode-SET_FUNCTION_ATTRIBUTE) instruction. This updates the run_in_dynamicscope to take account for it.

### DIFF
--- a/enaml/core/compiler_common.py
+++ b/enaml/core/compiler_common.py
@@ -61,9 +61,6 @@ _FUNC_DEF_NODES = (ast.Lambda,
                    ast.DictComp,
                    ast.SetComp)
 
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
I will try to think of something but I wonder what is your use case.

frmdstryr:
It appears the defaults can already access the dynamicscope so it's not a problem.

MatthieuDartiailh (approved):

PR #622, opened 2026-05-13
Patch d member after adding to class

frmdstryr:
I've been working on a port of atom more optimized for enaml-web. Due to how observation works in my port it requires that members be attached to a class before observers can be added. This is only change to enaml needed for my port to work. This makes it consistent with the other use of patch_d_member in DeclarativeMeta.

### DIFF
--- a/enaml/core/compiler_helpers.py
+++ b/enaml/core/compiler_helpers.py
@@ -154,8 +154,8 @@ def add_storage(node, name, store_type, kind):
     else:
         raise RuntimeError("invalid kind '%s'" % kind)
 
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Would you mind elaborating on what is specific to enaml-web workflow and the kind of optimization you implemented ?

frmdstryr:
Thanks!

There are a lot more child nodes getting added/removed than in a typical Qt application. The optimization is just a memory/speed tradeoff that switches the observer pool to use hashmaps instead of vectors.

PR #629, opened 2026-09-09
Fix f-string parsing bugs on Python 3.12+ (conversion specifiers, debug expressions, nested format specs)

### LINKED ISSUE #628: f-string parsing bugs on Python 3.12+: conversion specifiers, debug expressions (f'{x=}'), and nested format specs
I came across this bug when working on my code, and had a Claude AI agent work on this. Claude wrote up the issue and associated pull request. I reviewed and verified the work. 

## Description

`.enaml` files that use certain Python 3.12+ f-string features hit bugs in
enaml's PEG parser. What started as a single crash report (`!r`/`!s`/`!a`
conversion specifiers) turned out to be four related bugs on the same
`fstring_replacement_field`/`fstring_full_format_spec` code path, introduced
by PEP 701's tokenizer changes:

1. Any conversion specifier (`!s`, `!r`, `!a`) crashes the parser with an
   `AttributeError` instead of compiling.
2. A self-documenting debug expression (`f'{x=}'`) compiles, but silently
   loses its literal `"x="` prefix text from the resulting AST -- the
   rendered string is missing the label.
3. A debug expression combined with an explicit format spec
   (`f'{x=:>3}'`) wrongly gets an implicit `!r` conversion applied, when it
   should only default to `!r` when there's no format spec at all.
4. A format spec ending in a nested replacement field (`f'{x:{y}}'`)
   compiles, but its `format_spec` AST carries a spurious extra
   `Constant(value='')` node that CPython's own parser never produces.

None of these raise a `SyntaxError` for genuinely invalid syntax -- 1 is an
outright crash on **valid** input, and 2-4 are silent AST-correctness bugs
(the file compiles, but produces a wrong result).

## Minimal repros

Any `.enaml` file exercising these triggers them, e.g. in a function body:

```enaml
from enaml.widgets.api import Window

enamldef Main(Window):
    func describe(x):
        return f'{x!r}'
```

Each can also be reproduced directly against the parser, without a real
`.enaml` file, on Python 3.12+:

```python
from enaml.core.parser import parse
import ast

# Bug 1: crashes outright
parse("a = f'{x!r}'")

# Bug 2: compiles, but drops the 'x=' prefix
ast.dump(parse("a = f'{x=}'").body[0].ast.body[0])
# JoinedStr is missing Constant(value='x=') that ast.parse() includes

# Bug 3: compiles, but wrongly conversion=114 ('r') instead of -1
ast.dump(parse("a = f'{x=:>3}'").body[0].ast.body[0])

# Bug 4: compiles, but format_spec has a spurious trailing Constant('')
ast.dump(parse("a = f'{x:{y}}'").body[0].ast.body[0])
```

## Observed traceback (bug 1)

```
  ...
  File ".../enaml/core/parser/enaml_parser.py", line 5191, in fstring_conversion
    return self.check_fstring_conversion(conv_token, conv)
  File ".../enaml/core/parser/base_python_parser.py", line 224, in check_fstring_conversion
    if mark.lineno != name.lineno or mark.col_offset != name.col_offset:
       ^^^^^^^^^^^
AttributeError: 'TokenInfo' object has no attribute 'lineno'. Did you mean: 'line'?
```

Bugs 2-4 raise nothing -- they compile "successfully" to a subtly wrong AST,
which is arguably worse: nothing signals to a user that their f-string
compiled incorrectly.

## Root causes

- **Bug 1**: `check_fstring_conversion` (`base_python_parser.py`) checked
  adjacency between two `tokenize.TokenInfo` tokens using `.lineno`/
  `.col_offset`, attributes that exist on CPython's internal pegen
  tokenizer struct but **not** on Python's own `tokenize.TokenInfo` (which
  exposes `.start`/`.end` `(row, col)` tuples instead). The attribute
  access itself raised before the intended validation logic ever ran.
- **Bug 2**: `fstring_replacement_field`'s grammar action (`enaml.gram`)
  never emitted the leading `ast.Constant('x=')` node CPython injects for a
  debug expression -- it only used the `=` marker to pick the implicit
  conversion default, and otherwise dropped it.
- **Bug 3**: the same action's `b'r'[0] if debug_expr else -1` ternary
  didn't account for an explicit format spec being present alongside the
  debug marker.
- **Bug 4**: the tokenizer emits an `FSTRING_MIDDLE` token with an empty
  string for the zero-length literal-text run between a nested replacement
  field and whatever follows it (another field, or the closing `}`).
  `build_fstring_format_values` only special-cased a *lone* empty spec
  element (a bare `:` with nothing after), not one trailing other spec
  parts, so this degenerate token survived as a real, spurious AST node.

## Environment

- enaml: this repository's `main` branch (commit `51aaab0f`, "build(deps):
  bump codecov/codecov-action from 5 to 7 (#623)")
- Python: 3.14.6 (CPython, Windows); also affects 3.12 and 3.13 (verified
  on 3.12.7). Does **not** affect Python 3.10/3.11: below 3.12 there's no
  `FSTRING_START` token, so f-strings never reach this code path at all
  and go through the parser's legacy single-`STRING`-token path instead,
  which already handles all four cases correctly.

## Workaround

- Bug 1: avoid `!r`/`!s`/`!a` conversion specifiers in f-strings within
  `.enaml` files (use `str(x)`/`repr(x)` explicitly instead).
- Bugs 2-4: avoid debug expressions (`f'{x=}'`) and format specs containing
  a nested `{...}` replacement field in `.enaml` f-strings; there's no way
  to detect these silently-wrong cases short of avoiding the syntax.

## Fix

All four bugs, plus regression tests, are fixed across four commits on a
single accompanying pull request (they share enough of the same grammar
rule that separate PRs would conflict with each other on merge).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bburan:
Fixes #628. All was done by Claude and reviewed by me. Claude wrote up the description.

## Summary

`.enaml` files hit several bugs in the PEG parser's f-string handling on
Python 3.12+ (the PEP 701 tokenizer path): a crash on every conversion
specifier (`!s`/`!r`/`!a`), and two silent AST-correctness bugs that
compile without error but produce a wrong result (a debug expression's
`"x="` prefix text, and a spurious empty node in nested format specs).
This PR fixes all four bugs on that code path, one per commit, and adds
regression tests for each.

They're combined into one PR/branch rather than split up because they all
touch the same handful of lines in `fstring_replacement_field`'s grammar
action and its helpers -- split across separate PRs, the second to merge
would conflict with the first.

## Bug 1 -- `check_fstring_conversion` used the wrong token attributes

`enaml/core/parser/base_python_parser.py`, `check_fstring_conversion` (around
line 224), checked adjacency between the `!` mark token and the conversion
character token like this:

```python
if mark.lineno != name.lineno or mark.col_offset != name.col_offset:
```

Both `mark` and `name` are genuine `tokenize.TokenInfo` instances (this
method is called from the generated PEG parser,
`enaml/core/parser/enaml_parser.py:5191`, generated from
`enaml/core/parser/enaml.gram:1583`:
`conv_token="!" conv=NAME { self.check_fstring_conversion(conv_token, conv) }`,
whenever the parser encounters `!s`/`!r`/`!a` inside an f-string in a
`.enaml` file). `tokenize.TokenInfo` has no `.lineno`/`.col_offset` -- it
exposes `.start`/`.end` (each a `(row, col)` tuple) and `.line`. So this
line always raised:

```
AttributeError: 'TokenInfo' object has no attribute 'lineno'. Did you mean: 'line'?
```

for **every** conversion specifier, valid or not. The rest of this file
(`raise_syntax_error_known_location`, `raise_syntax_error_known_range`,
etc.) already branches on `isinstance(node, tokenize.TokenInfo)` and uses
`.start`/`.end` for that case -- this method was the one outlier that never
got that treatment.

The check's real intent is adjacency: the conversion character must come
immediately after the `!`, with nothing (not even whitespace) in between.
The correct translation using `TokenInfo`'s actual fields is:

```python
if mark.end != name.start:
```

## Bug 2 -- the conversion value that survives bug 1's fix is the wrong type

Fixing bug 1 alone still doesn't let a valid conversion specifier compile.
`fstring_conversion` is declared to return `int`
(`enaml.gram:1582`, `fstring_conversion[int]:`), and the `FormattedValue`
action that consumes it already treats the "no conversion" cases as plain
ints (`b'r'[0]` for a bare `=` debug expression, `-1` for none). But
`check_fstring_conversion` returned the raw `name` `TokenInfo`, and the
action then did:

```python
conversion.decode()[0] if conversion else (b'r'[0] if debug_expr else -1)
```

`TokenInfo` has no `.decode()`, so a syntactically valid, adjacent
conversion specifier still crashed -- same class of `AttributeError`, now
raised from `fstring_replacement_field` instead of
`check_fstring_conversion`.

Fixed by having `check_fstring_conversion` return `ord(s)` (the int the
grammar rule's own annotation already promised) and dropping the stray
`.decode()` call in the `enaml.gram` action, then regenerating
`enaml_parser.py` via `python -m enaml.core.parser.generate_enaml_parser`.
The regenerated file's only functional change is that one line.

## Bug 3 -- f-string debug expressions (`f'{x=}'`) lost their `x=` prefix

`fstring_replacement_field`'s grammar action never emitted the leading
`ast.Constant('x=')` node CPython injects for a self-documenting debug
expression. `f'{x=}'` compiled without error but produced a `JoinedStr`
missing that prefix, so the rendered string silently lost its `x=` label:

```
src: a = f'{x=}'
py:  JoinedStr(values=[Constant(value='x='), FormattedValue(...)])
en:  JoinedStr(values=[FormattedValue(...)])   # missing Constant('x=')
```

CPython's rule: the injected `Constant` is the raw source text between `{`
and whatever follows the `=` (the conversion `!`, the format spec `:`, or
the closing `}`), taken verbatim including interior whitespace, e.g.
`f'{ x = }'` -> `' x = '`.

Fixed by capturing the `{` token and, for the `=`? group, the token
immediately following it (peeked, not consumed, via a new
`fstring_debug_expr` sub-rule) -- giving enough position information to
slice that raw text out of the source without depending on which of
conversion/format/rbrace comes next. `fstring_replacement_field` now
returns `[Constant(...), FormattedValue(...)]` instead of a bare
`FormattedValue` when a debug expression is present; the `fstring_mid*`
and `fstring_format_spec*` loop results are flattened accordingly before
being used as a `JoinedStr`'s values (new helpers in
`base_python_parser.py`: `make_fstring_replacement_field`,
`_fstring_debug_prefix`, `flatten_fstring_parts`,
`build_fstring_format_values`).

Also fixes a second, closely related bug in the same conversion-default
logic: the implicit `!r` conversion for a bare debug expression was
applied even when an explicit format spec was present (`f'{x=:>3}'`
incorrectly got `conversion=114` instead of `-1`). CPython only defaults
to `!r` when there is no format spec at all -- with one, the value is
formatted directly, not repr'd first.

## Bug 4 -- spurious trailing empty `Constant` in nested format specs

A format spec ending in a nested replacement field, e.g. `f'{x:{y}}'`,
compiled without error but its `format_spec` `JoinedStr` carried an extra,
spurious `ast.Constant(value='')` CPython's own parser never produces:

```
src: a = f'{x:{y}}'
py:  format_spec=JoinedStr(values=[FormattedValue(value=Name('y'))])
en:  format_spec=JoinedStr(values=[FormattedValue(value=Name('y')), Constant(value='')])
```

Root cause: the tokenizer emits an `FSTRING_MIDDLE` token with an empty
string for the zero-length run of literal text between a nested
replacement field and whatever follows it (another field, or the closing
`}`). `fstring_format_spec` turns every `FSTRING_MIDDLE` into an
`ast.Constant` regardless of content, so this degenerate token became a
real, spurious element. `build_fstring_format_values` (introduced by the
bug 3 fix, as a rename/refactor of the prior inline ternary -- the
underlying filtering logic was unchanged) only special-cased a *lone*
empty element (a bare `:` with nothing after it), not one trailing other
spec parts, so it survived unfiltered.

Fixed by dropping every empty-string `Constant` from the flattened parts
before building the format spec's `JoinedStr`, regardless of position --
matching CPython, which never emits degenerate empty-text nodes into a
`JoinedStr`. A genuinely empty format spec (`f'{x:}'`) still correctly
produces an empty values list, now via the same general filter rather than
a separate special case.

Confined to `fstring_full_format_spec`'s nested-field handling; the
top-level `fstring_mid*` loop (the outer `JoinedStr`) was checked and does
not exhibit the same issue in practice -- `f'{x}{y}'` (two adjacent
replacement fields with no literal in between) already matches
`ast.parse()`.

## Testing

Added to `tests/core/parser/test_f_strings.py`:
- `test_f_string_conversion` -- parametrized over `s`/`r`/`a`, parses
  `f'{x!<conv>}'` and validates the resulting AST against `ast.parse()`.
  Runs on all supported Python versions (3.10+): below 3.12 it exercises
  the parser's legacy single-`STRING`-token path instead, which already
  handled conversions correctly, so no version skip is needed.
- `test_f_string_conversion_not_adjacent_to_mark` -- confirms `f'{x! r}'`
  (whitespace between `!` and the conversion character) still raises a
  `SyntaxError` with the intended message, rather than an
  `AttributeError`. Skipped below 3.12: enaml's own error wording is
  specific to this PEG code path, not reachable pre-3.12.
- `test_f_string_debug_expr` -- 6 cases (bare, whitespace, binop,
  format-spec, empty format-spec, multi-line), validated against
  `ast.parse()`. Skipped below 3.12 (`fstring_replacement_field` unreached).
- `test_f_string_nested_format_spec` -- 4 cases (single nested field, two
  adjacent nested fields, nested field followed by a literal, three
  adjacent nested fields), validated against `ast.parse()`. Skipped below
  3.12 (`fstring_full_format_spec` unreached).

Also spot-checked every case that's blocked by one bug while exercising
another, to confirm all four fixes compose correctly together:
`f'{x=!r}'` (debug expr + explicit conversion), `f'{x!r:>{w}}'`
(conversion + nested format spec), `f'{x=:{w}}'` (debug expr + nested
format spec) -- all match `ast.parse()` exactly after all four fixes are
applied together.

Ran the full parser suite (`python -m pytest tests/core/parser`) on Python
3.14 after all four fixes: 445 passed, 2 skipped (the 2 pre-existing skips
are an unrelated backslash-in-f-string test that only applies below 3.12).
Also ran the full test suite (`python -m pytest tests`): 4344 passed, 5
skipped, 0 failed.

Additionally spot-checked Python 3.11 (with `pegen>=0.3.0`, the version
this project requires) to confirm the legacy pre-3.12 path is unaffected
by any of the four fixes and already produces correct output for all of
them.

## Repro (before this fix)

```python
from enaml.core.parser import parse
parse("a = f'{x!r}'")
```

```
AttributeError: 'TokenInfo' object has no attribute 'lineno'. Did you mean: 'line'?
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DModRJvYu2WckV4FgT7vDP


MatthieuDartiailh:
Could you point Claude to pegen from which I extracted the parser to check if the bug need to fixed there too ?

bburan:
Looks like it, but most bugs were fixed by https://github.com/we-like-parsers/pegen/pull/118. Is the fix in Enaml now moot? The one pending bug is the `f'{x=}'`. See https://github.com/bburan/pegen/tree/fix/py312-fstring-bugs for the fix. If you think that should be included, I can issue a PR.

MatthieuDartiailh:
Thanks @bburan but I opened [pegen](https://github.com/we-like-parsers/pegen/pull/120) in the meantime. Once merged I will update the parser in enaml from the fixed pegen version.

bburan:
Closing. Will keep the issue open so that it can be marked off once Enaml is updated from Pegen.

